PR #29, opened 2020-01-02
update ios automatic codesigning from blog

rimanteval:


rimanteval:
https://blog.codemagic.io/automatic-code-signing-for-ios-that-doesnt-require-a-mac/

PR #43, opened 2020-01-21
Update clone URL information

helinanever:
Added information about how to find the clone URL and fixed a typo in examples.

### DIFF
--- a/content/getting-started/adding-apps-from-custom-sources.md
+++ b/content/getting-started/adding-apps-from-custom-sources.md
@@ -23,7 +23,7 @@ To add an app from custom source:
 
     {{< figure size="" src="../uploads/add-app-from-custom-source2.png" caption="" >}}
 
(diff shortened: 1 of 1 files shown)


content/getting-started/adding-apps-from-custom-sources.md
```diff
@@ -23,8 +23,8 @@ To add an app from custom source:
 
     {{< figure size="" src="../uploads/add-app-from-custom-source2.png" caption="" >}}
 
-    * Enter the **Repository URL** for cloning the repository. The URL should be in the following format:
-    `https://example.com/username/repo.git` or `git@example.com/username/repo.git`. 
+    * Enter the **Repository URL** for **cloning** the repository. You can find the URL from your repository on clicking **Clone** / **Clone or download**. Depending on the authentication method, choose either the HTTPS or SSH URL. The most common URL formats are as follows:
```
madispp: Currently, we do not support http auth for https:// repos, so 
> Depending on the authentication method
seems to be a bit misleading. Perhaps replace that sentence with something along the lines of:
> Depending on whether you're accessing a public or private repo

helinanever: Amended as you suggested.

PR #47, opened 2020-02-05
Rewrite iOS code signing

helinanever:
- updated screenshots
- added information about signing apps with extensions
- general editing


content/code-signing/ios-code-signing.md
```diff
@@ -4,21 +4,28 @@ title: iOS code signing
 weight: 1
 ---
 
-Code signing is required for installing your iOS app on real devices and publishing it to App Store. It enables to identify who developed the app and ensure that all the changes to the app come from you or your team.
+Code signing is required by Apple for integrating app services, installing your iOS app on real devices, and for uploading it to App Store Connect so that it can be distributed through TestFlight or App Store. It enables to identify who developed the app and ensure that all the changes to the app come from you or your team.
 
-To receive a signed `.ipa` file of your app on Codemagic, you need to set up code signing. When you build without code signing, you will receive only `Runner.app` that runs on simulators only.
+To receive a signed `.ipa` file of your app on Codemagic, you need to set up code signing. When you build without code signing, you will receive a file that runs on simulators only.
```
priitlatt: > you will receive a file that runs on simulators only.
This is not entirely true. We are not building for simulator architecture by default. The result is just a non-signed and non-packaged `.app` targeting iOS devices. Essentially useless.

To get a `.app`, which could be started on a simulator you'd have define `--simulator` flag in Flutter's iOS build arguments.

PR #53, opened 2020-02-13
adopt example for projects not in the root folder

mikhail-tokarev:


content/custom-scripts/load-firebase-configuration.md
```diff
@@ -11,10 +11,12 @@ Instead of committing the Firebase config files to your repository, you can uplo
 
         #!/usr/bin/env sh
 
-        set -e # exit on first failed commandset
+        set -e # exit on first failed command
+        
+        PROJECT_ROOT=$FCI_BUILD_DIR/myproject           # ADD YOUR PROJECT FOLDER HERE!!!
```
rimanteval: maybe would be better without the '!!!'

PR #67, opened 2020-03-25
Changing custom source repository URL information

evandropomatti:
I ended up waiting a couple of days to completely migrate our source code, so I think is useful for someone that ends up at the same problem.


rimanteval:
Thank you Evandro for this contribution, but we don't think this information will be helpful since we have already decided on implementing this feature. Therefore I will currently close this PR.

PR #71, opened 2020-04-01
Specify user roles and private key instructions

helinanever:


content/building/yaml.md
```diff
@@ -173,7 +173,7 @@ In order to use **automatic code signing** and have Codemagic manage signing cer
 
 * `CERTIFICATE_PRIVATE_KEY`
 
-  A RSA 2048 bit private key to be included in the signing certificate. Read more about it [here](https://help.apple.com/xcode/mac/current/#/dev1c7c2c67d). Tutorial for creating the 2048 bit RSA key - [link] (https://help.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent)
+  A RSA 2048 bit private key to be included in the [signing certificate](https://help.apple.com/xcode/mac/current/#/dev1c7c2c67d) Codemagic creates. This can be any existing or new private SSH key. You can find a tutorial for creating a new 2048 bit RSA key [here](https://help.github.com/en/github/authenticating-to-github/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent).
```
priitlatt, line 176: > This can be any existing or new private SSH key

I think this can be a little misleading since none of it has anything to do with SSH. It just has to be 2048 bit RSA private key, as opposed to, say 4096 bit DSA key.

PR #79, opened 2020-04-23
Update `POST /builds` parameters

priitlatt:


content/rest-api/builds.md
```diff
@@ -11,19 +11,30 @@ APIs for managing builds are currently available for developers to preview. Duri
...
-| ------------ | -------- | --------------- |
-| `appId`      | `string` | **Required.** The application identifier. |
-| `workflowId` | `string` | **Required.** The workflow identifier as specified in YAML file. |
-| `branch`     | `string` | **Required.** The branch name. |
+| **Name**      | **Type** | **Description** |
+| ------------- | -------- | --------------- |
+| `appId`       | `string` | **Required.** The application identifier. |
+| `workflowId`  | `string` | **Required.** The workflow identifier as specified in YAML file. |
+| `branch`      | `string` | **Required.** The branch name. |
+| `environment` | `object` | Optional. Specify environment variables and software versions to override values defined in workflow settings. |
```
mikhail-tokarev, line 19: should we add something like
`environment` object should contain at least one of `variables` or `softwareVersions` keys.

priitlatt: `environment` can be empty and both `variables` and `softwareVersions` can also be empty. All those are valid for `environment`:
```javascript
{}
{"variables": {}}
{"softwareVersions": {}}
{"variables": {}, "softwareVersions": {}}
```

PR #91, opened 2020-05-23
Update Codemagic API Docs

sbis04:
## Updates
* Add **Retrieve an application** call
* Add **Get build status** call
* Add `curl` request for all the calls

### DIFF
--- a/content/rest-api/applications.md
+++ b/content/rest-api/applications.md
@@ -5,62 +5,6 @@ weight: 2
 
 APIs for managing applications are currently available for developers to preview. During the preview period, the API may change without advance notice.
 
-## Retrieve applications list
-
-`GET /apps`
-
-Returns the full applications list present on your Codemagic account.
-
-### Curl request
-
-```bash
-curl -H "Content-Type: application/json" -H "x-auth-token: <API Token>" --request GET https://api.codemagic.io/apps
-```
-
-### Response
-
-```javascript
-{
-   "applications":[
-      {
-         "_id":"5d85eaa0e941e00019e81bc2",
-         "appName":"counter_flutter",
-         "branches":[
-            "master"
-         ],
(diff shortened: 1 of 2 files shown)


content/rest-api/applications.md
```diff
@@ -5,6 +5,40 @@ weight: 2
 
 APIs for managing applications are currently available for developers to preview. During the preview period, the API may change without advance notice.
 
+## Retrieve an application
```
mikhail-tokarev, line 8: currently, this endpoint is not public and we don't guarantee backward compatibility. Can you share more details about your use case to consider making the endpoint public?

content/rest-api/builds.md
```diff
@@ -56,8 +62,43 @@ APIs for managing builds are currently available for developers to preview. Duri
 }
 ```
 
+## Get build status
```
mikhail-tokarev, line 65: currently, this endpoint is not public and we don't guarantee backward compatibility. Can you share more details about your use case to consider making the endpoint public?

content/rest-api/applications.md
sbis04: Using this endpoint one can get all the application details like **workflowIds**, **branches**, etc., that can be helpful for running a new build using the endpoint `/builds` which has these required parameters.

content/rest-api/builds.md
sbis04: Currently, there is an endpoint for running new builds, but there is no way for tracking the build progress using the API. So, this would be helpful for tracking the builds and getting updates as it completes. And the user can perform some other operations as the build finishes.

content/rest-api/applications.md
orestesgaolin: I would love to see history of builds in the response, maybe together with artifacts links?

PR #94, opened 2020-05-27
Added links

Rrrrents:


content/building/yaml.md
```diff
@@ -360,3 +360,7 @@ For example, you can write a build script with Dart like this:
         app_id: '...'                     # App's unique identifier in App Store Connect
         apple_id: name@example.com        # Email address used for login
         password: Encrypted(...)          # App-specific password
+
+### Example
+
+If you want to know more about using YAML, there is a complete example of building a native iOS app with a Flutter module <a href="https://blog.codemagic.io/how-to-add-flutter-modules-to-native-ios-project-and-test-it-on-codemagic/" target="_blank" onclick="sendGtag('Link_in_docs_clicked','dynamic-workflows-with-codemagic-api')">here</a>. This article also describes how to use YAML to test this app on Codemagic CI/CD.
```
mikhail-tokarev, line 366: there is a copy-past in all google analytics events. It sends 'dynamic-workflows-with-codemagic-api' instead of proper names

content/code-signing/ios-code-signing.md
```diff
@@ -112,3 +112,7 @@ To export the provisioning profile:
 2. Navigate to **Certificates, Identifiers & Profiles > Profiles**.
 3. Select the provisioning profile you would like to export and click **Download**.
 4. Save it to have it ready.
+
+## Example
+
+If you are yearning for a more detailed example, Souvik Biswas has written a complete step-by-step <a href="https://blog.codemagic.io/distributing-native-ios-sdk-with-flutter-module-using-codemagic/" target="_blank" onclick="sendGtag('Link_in_docs_clicked','distributing-native-ios-sdk-with-flutter-module-using-codemagic')">overview about distributing a Native iOS project,</a> including both code signing and publishing it to App Store Connect.
```
helinanever, line 118: Let's not mention author names in docs, we don't want it to read too marketingy. Also, I'm not sure if iOS code signing doc is the right place for native iOS example. That particular reference would be more appropriate at the end of the yaml file doc or in the yaml template we will have about native iOS + Flutter module.

helinanever: Actually, we do have other blog posts that cover only the iOS code signing part, not sure how up to date they are, though.

content/rest-api/builds.md
```diff
@@ -5,6 +5,8 @@ weight: 3
 
 APIs for managing builds are currently available for developers to preview. During the preview period, the API may change without advance notice.
 
+If you have multiple similar workflows for the same project, you might be interested in using <a href="https://blog.codemagic.io/dynamic-workflows-with-codemagic-api/" target="_blank" onclick="sendGtag('Link_in_docs_clicked','dynamic-workflows-with-codemagic-api')">Dynamic Workflows</a> to make your life easier.
```
helinanever, line 8: I propose to have something more formal instead of "make your life easier" as it will be mainly companies/agencies that will need this feature, so I'm suggesting to use this: "If you have multiple similar workflows for the same project, you can configure your workflows dynamically using API calls, read more about it [here]." Also, since this information is not essential to the content of the doc, I'd add it as a note at the end of the page.

content/rest-api/overview.md
```diff
@@ -6,6 +6,8 @@ weight: 1
 
 Users can make calls to Codemagic REST API for better integration with other tools. Codemagic API can be accessed from `https://api.codemagic.io`. Any communication with it takes place over HTTPS. All data is sent and received as JSON.
 
+You can also use <a href="https://blog.codemagic.io/dynamic-workflows-with-codemagic-api/" target="_blank" onclick="sendGtag('Link_in_docs_clicked','dynamic-workflows-with-codemagic-api')">Dynamic Workflows</a> with the Codemagic REST API. It is especially useful when you are dealing with multiple similar workflows in the same project.
```
helinanever, line 9: I don't think we need two references to the same blog post within one topic.

PR #97, opened 2020-06-10
Fix api /builds POST docs 

BondarenkoStas:
POST /builds is done async and won't return result. Instead a user should find the started build from /builds endpoint.


Rrrrents:
LGTM

PR #113, opened 2020-07-16
update sample for native ios app

VeArnold:
**Description**

Updated the documentation of native ios apps in order to make sure the user is in the correct directory.


VeArnold:
yo bad change sry

PR #114, opened 2020-07-17
Update Native iOS and Native Android documentation for clarity

VeArnold:
**Description**

While by default when it comes to Native iOS and Native Android, we are actually in the right directory to execute the following commands, we still decided to change the code snippets a bit.

Due to the fact that we get many questions that could be avoided when people understood that changing directories and running the command should be done under the same script, we discussed with Mikhail that it could help if multiline scripts and changing directories would also be visible in places where users might see them better.

Using $FCI_BUILD_DIR also indicates to new users that we have environment variables that they can use to make their experience easier.

**Possible to-do's**
- make similar changes to elsewhere in the documentation
- find a place where to explain that each script is actually ran in a new shell
(this being the reason why the following would not work for example if we had a React Native app and wanted to build for android)

```
- cd android 
- ./gradlew build
```

### DIFF
--- a/content/yaml/building-a-native-ios-app.md
+++ b/content/yaml/building-a-native-ios-app.md
@@ -14,13 +14,19 @@ Codemagic uses the [xcode-project](https://github.com/codemagic-ci-cd/cli-tools/
 
 For building an unsigned iOS app (.app), you need to run the following command in the scripts section:
 
-    - xcodebuild build -workspace "MyXcodeWorkspace.xcworkspace" -scheme "MyScheme" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
+    - |
+      cd ios
+      xcodebuild build -workspace "MyXcodeWorkspace.xcworkspace" \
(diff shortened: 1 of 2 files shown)


content/yaml/building-a-native-ios-app.md
```diff
@@ -14,13 +14,19 @@ Codemagic uses the [xcode-project](https://github.com/codemagic-ci-cd/cli-tools/
 
 For building an unsigned iOS app (.app), you need to run the following command in the scripts section:
 
-    - xcodebuild build -workspace "MyXcodeWorkspace.xcworkspace" -scheme "MyScheme" CODE_SIGN_IDENTITY="" CODE_SIGNING_REQUIRED=NO CODE_SIGNING_ALLOWED=NO
+    - |
+      cd $FCI_BUILD_DIR
+      xcodebuild build -workspace "MyXcodeWorkspace.xcworkspace" \
+                       -scheme "MyScheme" \
+                       CODE_SIGN_INDENTITY="" \
+                        CODE_SIGNING_REQUIRED=NO \
```
mikhail-tokarev, line 22: formatting

PR #115, opened 2020-07-17
Fix build-universal-apk usage

BondarenkoStas:
There was an error, `build-universal-apk` has the option `--bundle`, not `--pattern` (that was changed).
Also would like to mention about possibilities to use glob patterns with cli tools.
Multiples apks can be generated if there are different .aab (matched by pattern), so better make plural.
Using no quotes or double quotes leads to some edge case complications, so it's better just to ask to use single quotes.

### DIFF
--- a/content/yaml/building-a-flutter-app.md
+++ b/content/yaml/building-a-flutter-app.md
@@ -18,7 +18,7 @@ Set up local properties
 
 ### Building universal .apk(s) from existing app bundle(s) with user-specified keys
 
-If your app settings in Codemagic have building Android App Bundles enabled, we will automatically include a script for generating a signed `app-universal.apk` during the YAML export. If you are creating a YAML file from a scratch, add the following script to receive those file(s):
+If your app settings in Codemagic have building Android App Bundles enabled, we will automatically include a script for generating a signed `app-universal.apk` during the YAML export. If you are creating a YAML file from a scratch, add the following script to receive the .apk file(s):
 
     - android-app-bundle build-universal-apk \
         --bundle 'project_directory/build/**/outputs/**/*.aab' \
--- a/content/yaml/building-a-flutter-app.md
+++ b/content/yaml/building-a-flutter-app.md
@@ -16,16 +16,18 @@ Set up local properties
 
     - flutter build apk --release
 
-### Building universal .apk from an existing app bundle with user-specified keys
+### Building universal .apk(s) from existing app bundle(s) with user-specified keys
 
-If your app settings in Codemagic have building Android App Bundles enabled, we will automatically include a script for generating a signed `app-universal.apk` during the YAML export. If you are creating a YAML file from a scratch, add the following script to receive that file:
+If your app settings in Codemagic have building Android App Bundles enabled, we will automatically include a script for generating a signed `app-universal.apk` during the YAML export. If you are creating a YAML file from a scratch, add the following script to receive those file(s):
 
     - android-app-bundle build-universal-apk \
-        --pattern 'project_directory/build/**/outputs/**/*.aab' \
+        --bundle 'project_directory/build/**/outputs/**/*.aab' \
         --ks /tmp/keystore.keystore \
         --ks-pass $CM_KEYSTORE_PASSWORD \
         --ks-key-alias $CM_KEY_ALIAS_USERNAME \
         --key-pass $CM_KEY_ALIAS_PASSWORD
+
+Please make sure to wrap the `--bundle` pattern in single quotes. If `--bundle` option is not specified, default glob pattern `**/*.aab` will be used.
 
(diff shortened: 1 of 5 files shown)


content/yaml/building-a-flutter-app.md
```diff
@@ -16,17 +16,19 @@ Set up local properties
 
     - flutter build apk --release
 
-### Building universal .apk from an existing app bundle with user-specified keys
+### Building universal .apk(s) from existing app bundle(s) with user-specified keys
 
-If your app settings in Codemagic have building Android App Bundles enabled, we will automatically include a script for generating a signed `app-universal.apk` during the YAML export. If you are creating a YAML file from a scratch, add the following script to receive that file:
+If your app settings in Codemagic have building Android App Bundles enabled, we will automatically include a script for generating a signed `app-universal.apk` during the YAML export. If you are creating a YAML file from a scratch, add the following script to receive those file(s):
```
Rrrrents, line 22: Change the ending for "... add the following script to receive the .apk file(s):"

content/yaml/runninglocally.md
```diff
@@ -29,22 +29,30 @@ To initialize keychain at system default keychain path with empty keychain passw
...
 To use the provisioning profile from `/path/to/profile.mobileprovision` in your Xcode project `/path/to/MyProject.xcodeproj` and generate an .ipa archive using `MyScheme` scheme, use [xcode-project](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/xcode-project/README.md#xcode-project) with the following command:
 
     xcode-project use-profiles --project /path/to/MyProject.xcodeproj --profile /path/to/profile.mobileprovision
     xcode-project build-ipa --project /path/to/MyProject.xcodeproj --scheme MyScheme
 
+Note that `--project` option can be either a path literal, or a glob pattern to match projects in working directory (Default is `**/*.xcodeproj`).
+
+`--profile` option can be a glob pattern as well (Default is `$HOME/Library/MobileDevice/Provisioning Profiles/*.mobileprovision`).
+
 ## Generate a universal APK with user specified keys from app bundle
```
Rrrrents, line 44: Would the following title work? (The thing is that we should follow the same logic and all other pages use .apk, not APK)
## Building universal .apk from an existing app bundle with user-specified keys

content/yaml/runninglocally.md
```diff
@@ -29,22 +29,30 @@ To initialize keychain at system default keychain path with empty keychain passw
...
 
-To build an APK from the app bundle `/path/to/my-app.aab` with keystore `/path/to/keystore.keystore`, `KEYSTORE_PASSWORD`, `KEY_ALIAS` and `KEY_PASSWORD`, use the [android-app-bundle](https://github.com/codemagic-ci-cd/cli-tools/tree/master/docs/android-app-bundle#android-app-bundle) tool:
+To build APK(s) from the app bundle(s) found with `/path/to/**/*.aab` glob pattern with keystore `/path/to/keystore.keystore`, `KEYSTORE_PASSWORD`, `KEY_ALIAS` and `KEY_PASSWORD`, use the [android-app-bundle](https://github.com/codemagic-ci-cd/cli-tools/tree/master/docs/android-app-bundle#android-app-bundle) tool:
```
Rrrrents, line 45: To build the .apk file(s)

content/yaml/runninglocally.md
```diff
@@ -29,22 +29,30 @@ To initialize keychain at system default keychain path with empty keychain passw
...
     android-app-bundle build-universal-apk \
-        --bundle 'path/to/my-app.aab' \
+        --bundle '/path/to/**/*.aab' \
         --ks /path/to/keystore.keystore \
         --ks-pass KEYSTORE_PASSWORD \
         --ks-key-alias KEY_ALIAS \
         --key-pass KEY_PASSWORD
 
+If `--bundle` option is not specified, default glob pattern `**/*.aab` will be used. Please make sure to wrap the pattern in single quotes.
+
```
Rrrrents, line 55: Do we say "default glob pattern" or should we be proper and say "default global pattern"?

content/yaml/building-a-flutter-app.md
BondarenkoStas: fixed in fdd4bc9a35859f468413142c28e2c13c8b1f18ff

content/yaml/runninglocally.md
BondarenkoStas: fixed in fdd4bc9

content/yaml/runninglocally.md
BondarenkoStas: Default global pattern is totally correct, but usually it's said as "glob patterns" as https://en.wikipedia.org/wiki/Glob_(programming), I think better to keep it that way

PR #123, opened 2020-07-23
change .keystore to .jks

VeArnold:
**Description**

During the last PR to change the Android code signing documentation, one minor detail was overlooked, as the file now generated is keystore_name.jks not keystore_name.keystore.

### DIFF
--- a/content/code-signing/android-code-signing.md
+++ b/content/code-signing/android-code-signing.md
@@ -31,7 +31,7 @@ You can create a keystore for signing your release builds with the Java Keytool 
 
     keytool -genkey -v -keystore keystore_name.jks -alias alias_name -keyalg RSA -keysize 2048 -validity 10000
 
-Keytool then prompts you to enter your personal details for creating the certificate, provide an alias for the key as well as provide passwords for the keystore and the key. It then generates the keystore as a file called **keystore_name.jks** in the directory you're in. The key is valid for 10,000 days.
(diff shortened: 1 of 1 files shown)


content/code-signing/android-code-signing.md
```diff
@@ -31,7 +31,7 @@ You can create a keystore for signing your release builds with the Java Keytool
 
     keytool -genkey -v -keystore keystore_name.jks -alias alias_name -keyalg RSA -keysize 2048 -validity 10000
 
-Keytool then prompts you to enter your personal details for creating the certificate, provide an alias for the key as well as provide passwords for the keystore and the key. It then generates the keystore as a file called **keystore_name.keystore** in the directory you're in. The key is valid for 10,000 days.
+Keytool then prompts you to enter your personal details for creating the certificate, provide an alias for the key as well as provide passwords for the keystore and the key. It then generates the keystore as a file called **keystore_name.jks** in the directory you're in. The key is valid for 10,000 days.
```
BondarenkoStas, line 34: in the command alias is already requested with -alias option, so probably shorten that sentence as 

> Keytool then prompts you to enter your personal details for creating the certificate and provide passwords for the keystore and the key ...

VeArnold:
> I think it's better to change it here to `keystore_name.keystore` because we use it with .keystore in other docs (yaml related). But In fact it doesn't even have to be .keystore or .jks, users can write whatever name and it will be fine.

I was not aware of the .keystore's in other documentation. I changed the example code snippet to .jks because that is how they have it on the Flutter website and also in one of our blog posts. However, whatever we go with, I think it should be the same everywhere, and as we can not edit the Flutter one, I think it is better to go with .jks everywhere.

PR #133, opened 2020-07-30
Improvement/rescan

BondarenkoStas:


content/getting-started/setting-up.md
```diff
@@ -54,12 +54,12 @@ After the build has finished successfully, you will immediately have **artifacts
 
 ## Multiple projects in one repository
 
-Codemagic supports monorepos and is able to detect multiple Flutter projects in a repository provided that each project has its `pubspec.yaml` file.
+Codemagic supports monorepos and is able to detect multiple Flutter projects in a repository or projects not in the repository root provided that each project has its `pubspec.yaml` file with `flutter` dependency.
```
Rrrrents, line 57: Flutter, with big L

content/getting-started/setting-up.md
```diff
@@ -54,12 +54,12 @@ After the build has finished successfully, you will immediately have **artifacts
...
 
-The first build is run for the project whose `pubspec.yaml` file was found first. After the first build, you can select the project for building from the **Project file path** dropdown in the Build section of app settings.
+Initially applications have only one project path `.` which means the root of the repository. We run scan during the first build and in case of multiple projects the first build is run for the project with highest number of platforms (presence of `android`, `ios` etc. directories). You can select the project for building from the **Project path** dropdown in the Build section of app settings. If you have changed your project location in the repository, renamed it's parent directory or added new projects, try **rescanning project paths**. The button is available on the right from **Project path** dropdown. It asks you which branch you want to search for projects and will update the project paths correspondingly to the content of the chosen branch.
```
Rrrrents, line 59: Initially, the only project path for every application is `.` — the root of the repository. We run scan during the first build and in case of multiple projects, the 

renamed its parent directory

the button -> There is a button for it on the right from the

It asks you, which branch you want to search for projects, and will update the project paths according to the content of the chosen branch.

content/getting-started/setting-up.md
BondarenkoStas: I think it's better to keep it with small one, in `pubspec.yaml` it looks like this 
```
dependencies:
  flutter:
    sdk: flutter
```

I meant about the name of dependency, not Flutter itself

content/getting-started/setting-up.md
BondarenkoStas: fixed in 9879df71d67396699e4738973f68a72374601496

PR #139, opened 2020-08-04
Update and rename how to run Fastlane lanes with Codemagic UI

martinjeret:
this is a general approach to running lanes with pre-build scripts in Codemagic UI configuration and does not have to be related to beta distribution. Moreover, the example is only related to UI configuration and should be explicitly said so.


madispp:
Other than the file change issue - this looks OK

PR #152, opened 2020-08-17
Fix ssh-keygen command for newer Mac OS versions

madispp:
User reported that the provided ssh-keygen command did not work for their certificates. This is because it generates an "OPENSSH" private key by default with newer versions of Mac OS.
Update command with an extra format specifier to force the PEM RSA format.

Source https://serverfault.com/a/941893


VeArnold:
another issue with this:
https://codemagicio.slack.com/archives/CEKE2KZ37/p1599201827281000

maybe should merge?

helinanever (approved):
LGTM

PR #163, opened 2020-08-31
update backend ip address

remarkov:
during the last major celery failure we recreated the GCP instance which lead to the change of IP address

NB: I wonder whether we need to send a notification to users about this change as those who already have the old IP in their firewalls may now experience connectivity issues.


VeArnold:
related to: https://github.com/NevercodeHQ/codemagic-support/issues/106

mikhail-tokarev (approved):
LGTM

PR #172, opened 2020-09-08
Update common-issues.md

VeArnold:
updated common issues,

removed 2FA authentication related error, as we do not authenticate users like that anymore, and added a common issue with App Store Connect set up.

### DIFF
--- a/content/troubleshooting/common-issues.md
+++ b/content/troubleshooting/common-issues.md
@@ -15,7 +15,7 @@ altool[xxx:xxx] *** Error: code -22020 (Unable to validate your application. We 
 
 The most common reason for the occurrence of this error message is that the user has not properly set up his/hers App-specific password and has used the Apple Developer account password instead.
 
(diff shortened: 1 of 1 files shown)


content/troubleshooting/common-issues.md
```diff
@@ -4,19 +4,18 @@ title: Common issues
 weight: 1
 ---
 
-## Two-factor authentication for Apple Developer Portal integration fails
+## iOS errors regarding creating authenticated sessions
```
helinanever, line 7: authenticated > authentication

content/troubleshooting/common-issues.md
```diff
@@ -4,19 +4,18 @@ title: Common issues
...
-
-If you see the error message above and are convinced you have entered the correct authentication code, you can try the following workarounds:
+```
+altool[xxx:xxx] *** Error: Unable to validate archive '/Users/builder/ipas/xxx'.
+altool[xxx:xxx] *** Error: code -22020 (Unable to validate your application. We are unable to create an authentication session.)
+```
 
- * Change your Apple Developer Portal password
- * Log out from all the devices except the one you want to use for receiving the authentication code
+The most common reason for the occurrence of this error message is that the user has not properly set up his/hers App-specific password and has used the Apple Developer account password instead.
```
helinanever, line 16: The most common reason for the occurrence of this error message is using Apple ID password instead of **app-specific password** in App Store Connect publishing settings.

helinanever: app-specific password can be linked there: https://support.apple.com/en-us/HT204397

helinanever (approved):

PR #177, opened 2020-09-16
Update title

martinjeret:
Current title of the documentation is misleading, because private dependencies are visible by default with OAuth authentication. We are trying to help developers access repositories outside of their repository that are not accessible by our OAuth scope. Hence rename title to accessing private dependencies and git submodules


helinanever:
This will be addressed in the reworked version of documentation.

PR #194, opened 2020-10-15
Code signing docs improvement

helinanever:
- Separated code signing for Android and iOS
- Added an example about signing with Gradle


content/code-signing-yaml/signing-ios.md
```diff
@@ -0,0 +1,109 @@
...
+
+  In **App Store Connect > Users and Access > Keys**, this is the **Key ID** of the key.
+
+* `APP_STORE_CONNECT_ISSUER_ID`
+
+  In **App Store Connect > Users and Access > Keys**, this is the **Issuer ID** displayed above the table of active keys.
+
+* `APP_STORE_CONNECT_PRIVATE_KEY`
+
+  This is the private API key downloaded from App Store Connect. Note that when encrypting files via UI, they will be base64 encoded and would have to be decoded during the build. Alternativey, you can encrypt the **contents** of the file and save the encrypted value to the environment variable.
```
BondarenkoStas, line 55: Just a thought, but maybe to give an example how to decode, like
```
  app-store-connect fetch-signing-files "io.codemagic.app"
    --type IOS_APP_DEVELOPMENT
    --create
    --private-key $(echo $APP_STORE_CONNECT_PRIVATE_KEY | base64 --decode)
```
or 
```
  echo $APP_STORE_CONNECT_PRIVATE_KEY | base64 --decode > decoded_private_key
  app-store-connect fetch-signing-files "io.codemagic.app"
    --type IOS_APP_DEVELOPMENT
    --create
    --private-key @file:decoded_private_key
```

But maybe to omit it right now and add a comprehensive section about yaml syntax and handling encryption and env vars and exit codes somewhere else, as it's applicable in other places as well.
I mean to emphasize that if it's a text file, user can encrypt just the content and won't need to decode it afterwards.

BondarenkoStas (approved):
I added a commit b4f5c0c56cf203790640e680cd83f77afaef155c, to fix the fact that used `CM_KEYSTORE_PATH` wasn't defined

content/code-signing-yaml/signing-ios.md
helinanever: I think it's a good idea to add that information under yaml syntax.

PR #210, opened 2020-11-10
Update signing-android.md

VeArnold:
When we export .yaml workflows that have Android code signing enabled, we export the alias as CM_KEY_ALIAS_USERNAME, but we recommend to use CM_KEY_ALIAS in documentation when setting up gradle, thus if users are exporting the .yaml, then this creates issues (and we're also telling them to set CM_KEY_ALIAS_USERNAME below).


helinanever (approved):
LGTM

PR #221, opened 2020-11-26
Update documentation storeFile environment var

wandriputra:
from generate download yaml file var name is FCI_KEYSTORE
...
        FCI_KEYSTORE: 
        FCI_KEY_PASSWORD: 
        FCI_KEY_ALIAS: 
...

### DIFF
--- a/content/code-signing/android-code-signing.md
+++ b/content/code-signing/android-code-signing.md
@@ -60,7 +60,7 @@ android {
     signingConfigs {
         release {
             if (System.getenv()["CI"]) { // CI=true is exported by Codemagic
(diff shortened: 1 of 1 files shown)


VeArnold:
The FCI_KEYSTORE_PATH and FCI_KEYSTORE are not exactly the same thing, and the action happens behind the scenes, however, for clarity, I think exporting the FCI_KEYSTORE_PATH alongside FCI_KEYSTORE, and using the path in the script instead of the tmp path will bring more clarity indeed, and I'll work on that soon.

VeArnold:
@wandriputra Thanks for the suggestion, we now export both FCI_KEYSTORE and FCI_KEYSTORE_PATH.

PR #224, opened 2020-12-07
[Docs - Getting Started] Remove branch_patterns as it is in different level

sbis04:
## Issue
The `branch_patterns` was defined at the same level as the `triggering`. It should be under the `triggering` as a **sub-property**. 
## Fix
Removed the `branch_patterns` property as here only a brief of the properties at the same level (under `my-workflow`) are shown.


helinanever (approved):
LGTM

PR #236, opened 2020-12-23
Change default build number to 1 instead of 0

jakecronin:
Building my android project locally caused a failure that the versionCode must be a positive integer greater than 0.


helinanever (approved):
LGTM

PR #238, opened 2021-01-11
Update publishing-to-app-store.md

myxiplx:
A section is needed here mentioning that you also need to register a device with your Apple developer profile, otherwise you get a build error if you simply follow the steps documented here to connect to Apple Connect and configure code signing.


craastad:
Hi. Thanks for reporting. I've recorded this PR as Issue https://github.com/codemagic-ci-cd/codemagic-docs/issues/299.

PR #240, opened 2021-01-18
Fix build number fetching instruction with yq v4

HugoHeneault:
See https://mikefarah.gitbook.io/yq/v/v4.x/upgrading-from-v3#reading-with-default-value :-)


mikhail-tokarev (approved):
LGTM

PR #253, opened 2021-02-03
add google play track env var

BondarenkoStas:
This is the env var which we respect, but it's not reflected in docs. Probably may be useful to somebody. Please rephrase in any way needed.

### DIFF
--- a/content/publishing-yaml/distribution.md
+++ b/content/publishing-yaml/distribution.md
@@ -70,7 +70,7 @@ publishing:
 
 ### Google Play
 
-Codemagic enables you to automatically publish your app to the `internal`, `alpha`, `beta` and `production` tracks on Google Play. In order to do so, you will need to set up a service account in Google Play Console and add the `JSON` key file to your Codemagic configuration file, see how to [set up a service account](../knowledge-base/google-play-api/).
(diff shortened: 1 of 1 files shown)


content/publishing-yaml/distribution.md
```diff
@@ -86,7 +86,7 @@ publishing:
 ```
 
 {{<notebox>}}
-The proper way to add your keys in `codemagic.yaml` is to copy the contents of the key file and [encrypt](../building/encrypting) it. Then add the encrypted value into the configuration file.
+You can override the publishing track specified in the configuration file using the environment variable `GOOGLE_PLAY_TRACK`. This is useful if you're starting your builds via [Codemagic API](../rest-api/overview/) and want to build different configurations without editing the configuration file.
 {{</notebox>}}
```
helinanever, line 90: The variable should also be documented in https://docs.codemagic.io/building/environment-variables/

BondarenkoStas: it's only for read-only env vars, ones that codemagic specifies itself and user can use

helinanever: OK, in this case we should just mention that the accepted values of the env variable are the same as in yaml config (I assume?)

BondarenkoStas: that's what i meant, track is taken from yaml config, but it can be overriden with the env var

helinanever (approved):
LGTM

PR #261, opened 2021-02-22
Update docs.md

martinjeret:
using latest version does not point to 11.7, because it is not the latest and .yaml users cannot select the version in build settings. You need to fix the version in order to use 11.7. Changes to docs are made accordingly.


priitlatt (approved):
:shipit:

PR #262, opened 2021-02-26
Updated google play api key docs with latest play store UI.

craastad:
Google play changed the name things, I updated the docs.

I realize one image is not in good proportion so, if it's really bad looking in the docs we can update it.

Old (left) vs. New (right) of updated images:
Screen Shot 2021-02-26 at 2 26 26 PM
Screen Shot 2021-02-26 at 2 26 23 PM
Screen Shot 2021-02-26 at 2 26 18 PM


helinanever (approved):
Can't verify it locally but the text changes look good to me.

craastad:
Images are autoscaled so the added image looks fine locally. :+1:
 [Archive.zip](https://github.com/codemagic-ci-cd/codemagic-docs/files/6049985/Archive.zip)

PR #272, opened 2021-03-10
Feature/big sur and linux versions

remarkov:
add versions report for macOS Big Sur image and Linux image (beta)

### DIFF
--- a/content/specs/versions-linux.md
+++ b/content/specs/versions-linux.md
@@ -3,7 +3,7 @@ description: A list of tools available out-of-the-box on Codemagic build machine
 title: Linux build machine specification
 aliases:
   - '../releases-and-versions/versions-linux'
-weight: 1
+weight: 4
 ---
 
 ## Hardware
--- a/content/specs/versions-linux.md
+++ b/content/specs/versions-linux.md
@@ -1,6 +1,6 @@
 ---
 description: A list of tools available out-of-the-box on Codemagic build machines.
-title: Linux build machine specification
+title: Linux (beta) build machine specification
 aliases:
   - '../releases-and-versions/versions-linux'
 weight: 4
--- a/content/specs/versions-linux.md
+++ b/content/specs/versions-linux.md
@@ -1,0 +1,48 @@
+---
+description: A list of tools available out-of-the-box on Codemagic build machines.
+title: Linux build machine specification
+aliases:
+  - '../releases-and-versions/versions-linux'
+weight: 1
+---
+
+## Hardware
+
+- Standard VM with Linux `4 vCPUs, 16 GB memory`
+- Premium VM with Linux `coming soon`
+
+## System
+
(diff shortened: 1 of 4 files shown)


content/specs/versions3.md
```diff
@@ -0,0 +1,148 @@
+---
+description: A list of tools available out-of-the-box on Codemagic build machines.
+title: macOS build machine specification (Xcode 12.5+)
+aliases:
+  - '../releases-and-versions/versions3'
+weight: 1
```
helinanever, line 6: weight: 3

content/specs/versions3.md
```diff
@@ -0,0 +1,148 @@
...
+- tar
+- unzip
+- wget
+- yarn `1.22.10`
+- yq
+- zip
+
+## Xcode 12.5 (12E5244e)
+
+This is the Xcode version used by default when you select `latest` in build settings. Other available versions are listed [here](#other-xcode-versions).
```
helinanever, line 54: Once we make this version available, it should be the edge, right? So I would change here "when you select 'edge'"

content/specs/versions3.md
```diff
@@ -0,0 +1,148 @@
...
+- Apple Watch Series 5 - 40mm
+- Apple Watch Series 5 - 44mm
+- Apple Watch SE - 40mm
+- Apple Watch SE - 44mm
+- Apple Watch Series 6 - 40mm
+- Apple Watch Series 6 - 44mm
+
+## Other Xcode versions
+
+- 12.4 (12D4e) `/Applications/Xcode-12.4.app`
```
helinanever, line 147: Are you sure we have 12.4 available on Big Sur image?

content/specs/versions-linux.md
```diff
@@ -0,0 +1,48 @@
+---
+description: A list of tools available out-of-the-box on Codemagic build machines.
+title: Linux build machine specification
+aliases:
+  - '../releases-and-versions/versions-linux'
+weight: 1
```
helinanever, line 6: weight: 4

content/specs/versions-linux.md
```diff
@@ -0,0 +1,48 @@
+---
+description: A list of tools available out-of-the-box on Codemagic build machines.
+title: Linux build machine specification
```
helinanever, line 3: I propose to add (beta) to the title

helinanever (commented):
Added some comments to address

content/specs/versions3.md
remarkov: why 3? as I understand that would make it after xcode-11 specs and we will have them in order "xcode-12", "xcode-11", "xcode-12.5". I think we should leave 1 here and shift the others to be below this one so the order will be "xcode-12.5", "xcode-12", "xcode-11".

helinanever: true, agree

helinanever (approved):
LGTM

PR #276, opened 2021-03-15
update versioning

martinjeret:
App Store, TestFlight and Google Play were at the end of section description, but should be in beginning to find it better


craastad (approved):
:+1:

PR #277, opened 2021-03-15
add tag query to get build from API doc

BondarenkoStas:
Now it's possible to query builds by tag, as well as to filter them from frontend, and filter for public dashboards


helinanever (approved):
LGTM

PR #279, opened 2021-03-17
Add automatic ios code signing youtube video

craastad:
Hugo has a nice [youtube shortcut](https://gohugo.io/content-management/shortcodes/#youtube) for embedding videos.

Screen Shot 2021-03-17 at 10 45 08 AM
Screen Shot 2021-03-17 at 10 52 03 AM

### DIFF
--- a/content/code-signing-yaml/signing-ios.md
+++ b/content/code-signing-yaml/signing-ios.md
@@ -21,6 +21,8 @@ Under the hood, we use [Codemagic CLI tools](https://github.com/codemagic-ci-cd/
 
 ## Automatic code signing
 
+{{< youtube 2kLi7Fe7q-0 >}}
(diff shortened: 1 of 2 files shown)


helinanever (approved):
Great stuff!

craastad:
Concerns brought up in Slack:

***Videos can go out of date***: Always a valid concern. If the video gets out of date and starts confusing users, we will remove it or recreate it (if users find it helpful).

***Mart suggested that videos should be at the bottom of text or section***: this is a valid suggestion. I aggree and disagree. Both the video and the text have the chance to go out of date. If users get engaged with the video they will likely continue on with code signing. If they don't want to watch the video they will read the text. In this context it makes sense right under the section it describes.

PR #287, opened 2021-03-23
Tweak example for FCI_CERTIFICATE decoding

fbartho:
This makes the code snippet slightly more flexible / ready to execute, such that people can have some certs with passphrases, and some without.

### DIFF
--- a/content/code-signing-yaml/signing-ios.md
+++ b/content/code-signing-yaml/signing-ios.md
@@ -128,10 +128,13 @@ scripts:
   - name: Set up signing certificate
     script: |
       echo $FCI_CERTIFICATE | base64 --decode > /tmp/certificate.p12
-      # when using a password-protected certificate
-      keychain add-certificates --certificate /tmp/certificate.p12 --certificate-password $FCI_CERTIFICATE_PASSWORD
-      # when using a certificate that is not password-protected
-      keychain add-certificates --certificate /tmp/certificate.p12
+      if [ "$FCI_CERTIFICATE_PASSWORD" = "" ]; then
+        # when using a certificate that is not password-protected
+        keychain add-certificates --certificate /tmp/certificate.p12
+      else
+        # when using a password-protected certificate
+        keychain add-certificates --certificate /tmp/certificate.p12 --certificate-password $FCI_CERTIFICATE_PASSWORD
+      fi
   - name: Set up code signing settings on Xcode project
     script: xcode-project use-profiles
   ... your build commands
--- a/content/code-signing-yaml/signing-ios.md
+++ b/content/code-signing-yaml/signing-ios.md
@@ -128,7 +128,7 @@ scripts:
   - name: Set up signing certificate
     script: |
       echo $FCI_CERTIFICATE | base64 --decode > /tmp/certificate.p12
-      if [ "$FCI_CERTIFICATE_PASSWORD" = "" ]; then
+      if [ -z ${FCI_CERTIFICATE_PASSWORD+x} ]; then
         # when using a certificate that is not password-protected
         keychain add-certificates --certificate /tmp/certificate.p12
       else


fbartho:
possibly we should add quotes too, in case the passphrase has spaces in it?

content/code-signing-yaml/signing-ios.md
```diff
@@ -128,10 +128,13 @@ scripts:
   - name: Set up signing certificate
     script: |
       echo $FCI_CERTIFICATE | base64 --decode > /tmp/certificate.p12
-      # when using a password-protected certificate
-      keychain add-certificates --certificate /tmp/certificate.p12 --certificate-password $FCI_CERTIFICATE_PASSWORD
-      # when using a certificate that is not password-protected
-      keychain add-certificates --certificate /tmp/certificate.p12
+      if [ "$FCI_CERTIFICATE_PASSWORD" = "" ]; then
```
priitlatt, line 131: I'm afraid this is not the correct way to perform this check. In case the variable `FCI_CERTIFICATE_PASSWORD` is not set then the condition is invalid both in `bash` and `zsh` and the else clause will be executed, which is not desired:
```bash
# in bash
$ if [ ${pw} = "" ]; then echo "password is not set"; else echo "passowrd is set to '$pw'"; fi
bash: [: =: unary operator expected
passowrd is set to ''
```
```bash
# in zsh
$ if [ ${pw} = "" ]; then echo "password is not set"; else echo "passowrd is set to '$pw'"; fi
zsh: parse error: condition expected: =
passowrd is set to ''
```

Better way to actually check whether the password is given or not is to use the condition
```bash
[ -z ${FCI_CERTIFICATE_PASSWORD+x} ]
```

See this [SO thread](https://stackoverflow.com/a/13864829) for more in depth explanation.

priitlatt (changes requested):

content/code-signing-yaml/signing-ios.md
fbartho:
```suggestion
      if [ -z ${FCI_CERTIFICATE_PASSWORD+x} ]; then
```

like so?

fbartho: Note: my original code had extra double quotes that weren’t present in the sample code that you tested @priitlatt!

I believe my original code is safe on several shells, but I have no problem changing to your suggestion.

priitlatt: Indeed, I managed to miss that you had wrapped the variable in quoetes. Nonetheless, even when using double quotess, the direct comparison does not distinguish unset variable from variable with empty value:
```bash
export password=''  # Set $password to empty value
if [ "${password}" = "" ]; then echo "password is not set"; else echo "passowrd is set to '$pw'"; fi
```
```bash
$ unset password  # Make sure that $password is not defined
$ if [ "${password}" = "" ]; then echo "password is not set"; else echo "passowrd is set to '$pw'"; fi
password is not set
```
However the `-z` check differentiates between the two:
```bash
export password=''  # Set $password to empty value
$ if [ -z ${password+x} ]; then echo "password is not set"; else echo "passowrd is set to '$pw'"; fi
passowrd is set to ''
```
```bash
$ unset password  # Make sure that $password is not defined
$ if [ -z ${password+x} ]; then echo "password is not set"; else echo "passowrd is set to '$pw'"; fi
password is not set
```

While it may seem like an extraneous detail, it has a significant difference for the underlying operations. Namely, one can "encrypt" a certificate with empty password, and then this certificate needs to be treated as password protected. Hence it would require the environment variable to be set up as `FCI_CERTIFICATE_PASSWORD=''`, and the script should accept the value from this variable.

priitlatt (approved):

content/code-signing-yaml/signing-ios.md
fbartho: I certainly wouldn't do that (because why⁇), but obviously for your code-snippets you probably want the better solution that you suggested :)

PR #296, opened 2021-04-07
update docs

martinjeret:
Update information about Google Play app being in draft status and service account error


mikhail-tokarev (approved):
LGTM

craastad (approved):
It's probably should be in a troubleshooting section, but it's fine for now. :+1:

PR #306, opened 2021-04-14
Create Firebase Crashlytics dsym uploading example

martinjeret:
We've gotten many requests how to do this. Would be nice to document it


craastad (approved):
Ok :+1:

PR #323, opened 2021-05-03
fix wrong command to require buildnumber by using yq

remonh87:
The output of the command  in the current docs is incorrect for the following pubspec entry: `version: 1.0.0+1` . When running `yq e .version pubspec.yaml`  the result will be `1.0.0.+1` which results in an error that the buildnumber needs to be an integer. The changes below provides a working example.

### DIFF
--- a/content/building/build-versioning.md
+++ b/content/building/build-versioning.md
@@ -20,12 +20,12 @@ Here are some examples how you can increment the app version using Codemagic's r
 --build-name=1.0.0 --build-number=$BUILD_NUMBER
 ```
 
-## Fetching build number from pubsec.yaml
+## Fetching build name and build number from pubsec.yaml
 
 Add the following build argument:
 
 ```bash
---build-number=$(yq e .version pubspec.yaml)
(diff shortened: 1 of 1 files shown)


content/building/build-versioning.md
```diff
@@ -20,12 +20,12 @@ Here are some examples how you can increment the app version using Codemagic's r
...
 ```
 
-## Fetching build number from pubsec.yaml
+## Fetching build name and build number from pubsec.yaml
 
 Add the following build argument:
 
 ```bash
---build-number=$(yq e .version pubspec.yaml)
+--build-name=$(yq e .version pubspec.yaml | cut -d "+" -f1) --build-number=$(yq e .version pubspec.yaml | cut -d "+" -f2)
```
priitlatt, line 28: In case version does not have the `+int` modifier, then both build name and build version will be the same `x.y.z`, which will result in invalid build number, doesn't it?

remonh87: yes so maybe it is good to add the example as well. The[ official flutter docs](https://flutter.dev/docs/deployment/android#updating-the-apps-version-number) describe that `x.y.z` is used for version and `+w` is used for buildnumber.

PR #339, opened 2021-05-12
more noteboxes that vars should be decoded

BondarenkoStas:
I'm so used that during encryption files are base64 encoded that i'm decoding them without paying attention to it, or encode the content itself. @craastad  it would be cool if you can add something from yourself to catch more attention to it.

### DIFF
--- a/content/building/build-versioning.md
+++ b/content/building/build-versioning.md
@@ -79,7 +79,11 @@ It is recommended to create a dedicated App Store Connect API key for Codemagic 
 
 - `APP_STORE_CONNECT_PRIVATE_KEY`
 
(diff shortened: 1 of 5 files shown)


content/building/build-versioning.md
```diff
@@ -79,7 +79,11 @@ It is recommended to create a dedicated App Store Connect API key for Codemagic
 
 - `APP_STORE_CONNECT_PRIVATE_KEY`
 
-  This is the private API key downloaded from App Store Connect. Note that when encrypting files via UI, they will be base64 encoded and would have to be decoded during the build. Alternativey, you can encrypt the **contents** of the file and save the encrypted value to the environment variable.
+  This is the private API key downloaded from App Store Connect.
+
+{{<notebox>}}
+  Note that when encrypting files via UI, they will be base64 encoded and would have to be decoded during the build. Alternativey, you can encrypt the **contents** of the file and save the encrypted value to the environment variable.
```
craastad, line 85: This is what I'm doing but still running into base64 issues.

PR #346, opened 2021-05-17
rephrase "stop build if tests fail" description

BondarenkoStas:
The previous description implies that a build will stop immediately if a test fails, but the checkbox is related to the whole test step, and all the enabled tests will run.

Inspired by https://codemagicio.slack.com/archives/CEKE2KZ37/p1621004312408700


helinanever (approved):
LGTM

PR #349, opened 2021-05-18
add --warn-only flag

nevercode-kevin:
There have been several occasions when React Native projects contain a dependency that also contains an .xcodeproj and Codemagic tries to find a provisioning profile to match the project which in turn causes the build to fail. Adding the --warn-only flag will just show a warning instead of making the build fail unnecessarily.


craastad:
Duplicated by https://github.com/codemagic-ci-cd/codemagic-docs/pull/351#pullrequestreview-662851367

PR #358, opened 2021-05-19
Add Algolia search

craastad:
I added [Algolia search](https://www.algolia.com/products/search/) using their [javascript client](https://github.com/algolia/algoliasearch-client-javascript).

TODO
- [x] decide if console.log(...) is OK so we can check Algolia vs. Fuse search results if need be (remove console.log)
- [x] decide best way to leave existing code so we can revert back to it in a week or two if we're unhappy with Algolia (remove fuse code)
- [ ] when deploying codemagic docs we need to set up a trigger to update the `codemagic_docs` index in Algolia ([API call](https://www.algolia.com/doc/api-reference/api-methods/replace-all-objects/?client=javascript)) with the up to date `index.json` file

Algolia "Flutter Android Signing" search
Screen Shot 2021-05-19 at 15 54 06

Existing "Flutter Android Signing" search
Screen Shot 2021-05-19 at 15 47 19

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -98,10 +98,8 @@ const updateUrl = (query) =>
 const updateResults = async (query) => {
     let result, algoliaResult
     try {
-        result = getResults(query)
-        console.log("fuse: ", result)
+        // result = getResults(query)
         algoliaResult = await getAlgoliaResults(query)
-        console.log("algolia: ", algoliaResult)
     } catch (error) {
         result = error
     }
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -290,12 +290,8 @@ const getAlgoliaResults = (query) =>
     query
         ? algolia
             .search(`'${query}`, {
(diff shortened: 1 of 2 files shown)


assets/js/search.js
```diff
@@ -1,5 +1,7 @@
 let fuse
 
+const algolia = algoliasearch("27CIRMYZIB", "7e88305c04e90188508daa6c89e5f4df").initIndex("codemagic_docs");
```
craastad, line 3: These are perfectly fine to be public, see the [InstaSearch.js](https://codesandbox.io/embed/github/algolia/doc-code-samples/tree/master/InstantSearch.js/getting-started) demo.

mikhail-tokarev (approved):
🚀

VladyslavBondarenko (approved):
I'll address index update in https://github.com/codemagic-ci-cd/codemagic-docs/pull/362, it can be deployed separately

PR #359, opened 2021-05-19
Updated release notes

craastad:


mikhail-tokarev (approved):
🚀 
ps: I think newsletter is better and we can remove this page

PR #362, opened 2021-05-20
Algolia search index update

VladyslavBondarenko:
Update search index on deploy using [algolia-indexing](https://www.npmjs.com/package/algolia-indexing).

Addition to https://github.com/codemagic-ci-cd/codemagic-docs/pull/358


package.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "codemagic-docs",
+    "version": "1.0.0",
+    "scripts": {
+        "algolia": "node algolia.js"
+    },
+    "dependencies": {
```
mikhail-tokarev, line 7: probably dev dependencies should be ok

.gitignore
```diff
@@ -1,3 +1,4 @@
 /public/
 /resources/
 .DS_Store
+*node_modules/
```
mikhail-tokarev, line 4: add full path

.github/workflows/deploy.yml
```diff
@@ -18,3 +18,7 @@ jobs:
       - run: curl --location https://github.com/gohugoio/hugo/releases/download/v0.77.0/hugo_extended_0.77.0_Linux-64bit.tar.gz | tar -vxzO hugo > hugo && chmod a+x hugo
       - run: ./hugo --environment production --minify
       - run: ./hugo deploy --environment production --invalidateCDN true --maxDeletes -1
+
+      - run: yarn install
```
mikhail-tokarev, line 22: should it be before `./hugo deploy`?

mikhail-tokarev (commented):
added some comments

.gitignore
mikhail-tokarev: should be only one node_modules otherwise there is some problem

.github/workflows/deploy.yml
VladyslavBondarenko: do you mean only `yarn install` or together with `yarn run algolia`?

mikhail-tokarev (approved):
LGTM

PR #363, opened 2021-05-21
New add app flow

helinanever:
Covers the new flow for adding applications.
There are substantial changes in https://docs.codemagic.io/getting-started/adding-apps-from-custom-sources/ and https://docs.codemagic.io/getting-started/signup/.
Does not cover the changes related to Applications dashboard.

[do not merge before the feature is public]


craastad (approved):
Good job 🎉

PR #366, opened 2021-05-24
Provide users with scripts that are compatible with both MacOS and current Linux VMs

bean5:
I was using code from https://docs.codemagic.io/knowledge-base/add-a-git-tag-with-app-version/. That started failing when I started using Linux VMs.

It appears that https://docs.codemagic.io/publishing/beta-deployment-with-fastlane/ would suffer from the same issue. There might be a way to support zsh instead or sh directly, but I wasn't sure what you would prefer so I defaulted to bash to just get it done.


artemii-yanushevskyi (approved):

artemii-yanushevskyi:
Thank you for contributing, Michael

bean5:
Thanks for thanking. I help where I can, @artemii-yanushevskyi .

PR #378, opened 2021-06-03
Minor correction in SSH keys docs

remarkov:
We actually show the message about adding the SSH keys to the agent during the "Fetching app sources" step, not during "Preparing build machine"

image


helinanever (approved):
LGTM

PR #381, opened 2021-06-07
Add info about templates

martlaub:


mikhail-tokarev:
will close without merge since we are not going to release templates to everyone

PR #385, opened 2021-06-14
ensure keystore has JKS format

icarusdust:


mikhail-tokarev:
see https://codemagicio.slack.com/archives/CEKE2KZ37/p1623685985110100 for more information

VeArnold (changes requested):
I think it might be better to keep the formatting snake_case and a bit more meaningful as per other examples. Any particular reason to renaming the jks file to upload-keystore? Perhaps keep it keystore_name or change it to upload_keystore. As for the alias, perhaps revert back to alias_name too?

icarusdust:
Sure thing. Committing the changes as suggested. No, there is no particular reason for renaming to upload-keystore.

VeArnold (approved):
LGTM, thanks for your contribution! :)

PR #389, opened 2021-06-17
Add examples of using flutter test integration_test

VeArnold:


content/testing-yaml/testing.md
```diff
@@ -33,6 +33,18 @@ scripts:
 
 The `integration_test` dependency allows you to run integration tests on a real device or emulator. Android application tests can be run on an Android emulator, iOS application tests can be run on an iOS simulator, and web application tests can be run on a web browser driver.
 
+**Tip:** It is also possible to use `flutter test` to run integration tests using the integration_test package. When using `flutter test` instead of `flutter drive` it is possible to generate machine readable output using the `--machine` flag, hence the results can be displayed in the UI. Just include the `test_report` field with a glob pattern matching the test result file location:
```
helinanever, line 36:
```suggestion
**Tip:** It is also possible to use `flutter test` to run integration tests using the integration_test package. When using `flutter test` instead of `flutter drive`, it is possible to generate machine-readable output using the `--machine` flag, and the results can be displayed in the UI. Just include the `test_report` field with a glob pattern matching the test result file location:
```

helinanever (approved):
LGTM

PR #402, opened 2021-06-30
Update environment-variables.md

icarusdust:
$FCI_BUILD_STEP_STATUS, explicitly pointing out that it is for UI, not yaml.


mikhail-tokarev (approved):
nip: maybe we can provide a link to an example of how to achieve the same result in yaml file

PR #409, opened 2021-07-05
More detailed information about pricing

nevercode-kevin:
More details about instance types and pricing added.


config.yml
```diff
@@ -78,8 +78,8 @@ menu:
     - identifier: teams
       name: Teams
       weight: 7
-    - identifier: billing
-      name: Billing
+    - identifier: billing-pricing
```
helinanever, line 81: We shouldn't change the identifier that easily, it will break links from the app. We could use page aliases but I think in this case it makes sense to only change the name of the category and not the name of the folder.

content/billing-pricing/pricing.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Pricing
+weight: 2
+---
+
```
helinanever, line 5: Should there be an intro about the free tier and what's included as well to explain the whole model?

content/billing-pricing/pricing.md
```diff
@@ -0,0 +1,105 @@
+---
+title: Pricing
```
helinanever, line 2: For the record, currently there are no links from Billing page to this page, the Billing page refers to the pricing page on our landing for pricing related info.

content/billing-pricing/pricing.md
```diff
@@ -0,0 +1,105 @@
...
+weight: 2
+---
+
+## Pay as you go pricing
+
+Pay as you go pricing allows you to pay for only what you use. 
+
+Currently it is only possible to run one build at a time with pay as you go.
+
+### Instance Types
```
helinanever, line 12: I feel like the instance types information would be more suited to this page: https://docs.codemagic.io/specs/machine-type/

content/billing-pricing/pricing.md
helinanever: OK, I can see this is addressed later in the file.

content/billing-pricing/pricing.md
nevercode-kevin: Instance types should be shown on this page so people can see what the price relates to without having to jump to another page.

config.yml
nevercode-kevin: reverted to original name

helinanever (approved):
LGTM

PR #413, opened 2021-07-09
Update encrypting.md

icarusdust:
encryption tool in the Workflow Editor, otherwise it could be vague what tool we are referring to

### DIFF
--- a/content/building/encrypting.md
+++ b/content/building/encrypting.md
@@ -29,7 +29,7 @@ On macOS, in order to base64 encode a file and copy the contents to your clipboa
   ```bash
   cat your_file_name.extension | base64 | pbcopy 
   ```
-After the command has been run, you can paste the string from your clipboard to our encryption tool and then save the encrypted string to an environment variable.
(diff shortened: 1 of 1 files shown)


content/building/encrypting.md
```diff
@@ -29,7 +29,7 @@ On macOS, in order to base64 encode a file and copy the contents to your clipboa
   ```bash
   cat your_file_name.extension | base64 | pbcopy 
   ```
-After the command has been run, you can paste the string from your clipboard to our encryption tool and then save the encrypted string to an environment variable.
+After the command has been run, you can paste the string from your clipboard to our encryption tool in the Workflow Editor and then save the encrypted string to an environment variable.
```
helinanever, line 32: It's not really the workflow editor as the same interface is available in both yaml view and in the app settings for Flutter apps, and workflow editor strictly refers to the UI settings of Flutter apps. Perhaps use "in the Codemagic UI" instead?

icarusdust: Agreed, Codemagic UI sounds a lot better, going to submit it

helinanever (approved):
LGTM

PR #418, opened 2021-07-15
Specify android paths

VeArnold:
Flutter for some reason copies the identical apk to multiple folders - thus with the paths we recommend currently, we export three artefacts with the same name, which is confusing for the users (the md5 checksum is the same - meaning they literally are identical). The change to export more specific paths is already deployed live. 

Screenshot 2021-07-15 at 15 05 15


helinanever (approved):
LGTM

PR #423, opened 2021-07-22
Update firebase-app-distribution.md

craastad:
Updated Firebase app distribution to explicitly reference aliases.


craastad:
what's there is  close enough

PR #449, opened 2021-08-18
Improvement: move extra deployment options to Knowledge base

artemii-yanushevskyi:
These articles will also be useful for YAML based configuration.
Made sure that URLs will not break


helinanever (approved):
LGTM

PR #453, opened 2021-08-23
add optional teamId param to new app API

remarkov:


helinanever (approved):
LGTM

mikhail-tokarev:
should we describe how to find team id?

remarkov:
@mikhail-tokarev don't think so, at least not in this section of docs. probably need to add extend API description with other endpoints, e.g. to retrieve user/team data but that should be a different task.

PR #461, opened 2021-08-30
Update building-a-flutter-app.md

icarusdust:
When generating .apk and .aab, they seem to be located incorrectly. Sample Flutter projects are also set up in this way.

### DIFF
--- a/content/yaml-quick-start/building-a-flutter-app.md
+++ b/content/yaml-quick-start/building-a-flutter-app.md
@@ -134,8 +134,8 @@ workflows:
         script: |
           cd . && flutter build apk --release --build-name=1.0.0 --build-number=$(($(google-play get-latest-build-number --package-name "$PACKAGE_NAME" --tracks="$GOOGLE_PLAY_TRACK") + 1))
     artifacts:
-      - build/**/outputs/apk/**/*.apk
-      - build/**/outputs/bundle/**/*.aab
+      - build/**/outputs/**/*.apk
+      - build/**/outputs/**/*.aab
       - build/**/outputs/**/mapping.txt
       - flutter_drive.log
     publishing:


VeArnold:
Hey - could you elaborate or provide some project where this is an issue? The paths work perfectly well with default Flutter projects.

Thanks

icarusdust:
Hey, this is what I see in the logs: ✓  Built build/app/outputs/flutter-apk/app-release.apk (15.3MB).
In this case, - build/**/outputs/apk/**/*.apk does not match the built apk path. Am I mistaken ? Therefore, build/**/outputs/**/*.apk should be used ?

icarusdust:
seems like github ignore star signs ))

VeArnold:
I think this change was distinctively added due to the fact that Flutter produced multiple artefacts, which got sent to the user as well then, if the path was too ambiguous (the artefacts themselves were identical). If you confirm that you can not get any artefacts with these paths, let me know.

icarusdust:
I am checking Flutter generated artifacts with some other users and my personal projects, and all of them have distinctive paths. Most are build/app/outputs/flutter-apk/app-release.apk.

VeArnold:
Yes, I get that, but Flutter duplicates the artefacts as for as I know, I am asking if you have a project/build where these paths have not produced and artefact in publishing.

I am pretty sure these paths:
      - build/**/outputs/**/*.apk
      - build/**/outputs/**/*.aab
 
will produce three identical artefacts due to some flutter magic - you can test.

icarusdust:
And if Flutter generates multiple artifacts, should not we use it as build//outputs//.apk. Besides, all of our sample projects are set up with this path.

VeArnold:
The issue that was solved was the fact that users were getting three identical artefacts to their e-mail, and it confused them. I would rather see you update the example projects, than revert this.

icarusdust:
I could reproduce the issue with 3 identical artefacts

Issue #445, opened 2021-08-16
Testing: docs are missing the command to wait for device when launching the emulator


helinanever:
Running Flutter integration tests from yaml fail because the build doesn't wait for the emulator to be launched.
Adding the `adb wait-for-device ` command fixes the issue.

```
- name: Launch emulator
  script: |
    flutter emulators --launch emulator
    adb wait-for-device 
```
The same command can be documented for other project types too.

icarusdust:
I would assume that adding **adb wait-for-device** is not only Flutter integration test related but also for all the platforms and it should be added no only under Flutter section but somewhere here: https://docs.codemagic.io/yaml-testing/testing/

icarusdust:
well, not actually because the other examples already include it, so it should be Flutter specific

PR #464, opened 2021-08-31
Update testing.md

icarusdust:


content/yaml-testing/testing.md
```diff
@@ -46,6 +46,15 @@ scripts:
     test_report: test-results/flutter.json
 ```
 
+If running Flutter integration tests from yaml fails because the build doesn't wait for the emulator to be launched, then adding **adb wait-for-device** command could fix the issue:
```
helinanever, line 49: Just wondering if this command should be included in the example below by default?

icarusdust: It was added to the default script in React Native integration test section on the same page. I believe we can also add it to the Flutter the same way rather than making it as a separate paragraph as I did in the PR.

icarusdust: For example:
flutter emulators --launch apple_ios_simulator  &        # for android use: flutter emulators --launch emulator
adb wait-for-device

content/yaml-testing/testing.md
```diff
@@ -46,19 +46,11 @@ scripts:
     test_report: test-results/flutter.json
```
helinanever: should we add the command in the example below too?

helinanever (approved):
LGTM

icarusdust:
Fixes https://github.com/codemagic-ci-cd/codemagic-docs/issues/445

PR #469, opened 2021-09-02
Feature/dart code metrics

VeArnold:


content/yaml-testing/dart-code-metrics.md
```diff
@@ -4,6 +4,10 @@ description: How to run static code analysis with Dart Code Metrics
 weight: 3
 ---
 
+{{<notebox>}}
+For the following instructions to work with Codemagic, Dart version 2.12.0 or higher is required.
```
helinanever, line 8: Why do we mention Dart version here and not the Flutter version?

VeArnold: You theoretically don't have to have a Flutter project to run dart code metrics (or use dart files) - for UI it's presumed it is a Flutter project.

helinanever: I mean we don't have a good way to configure the Dart version and afaik we don't have the standalone Dart SDK anymore, whatever Dart version is used comes from the Flutter SDK.

helinanever: Maybe an alternative would be to mention both Dart and Flutter versions?

helinanever (approved):
LGTM

PR #470, opened 2021-09-02
rename content directories according to menu categories

VladyslavBondarenko:
Breadcrumbs will be formed from the URL path. With current directory names article `Advanced configuration > Build versioning` would be shown in breadcrumbs as `Configuration > Build versioning`, because the path is `/configuration/build-versioning/`.

When directories are renamed, breadcrumbs will match menu items.

Added aliases to renamed articles for backward compatibility.


VladyslavBondarenko:
I'll try to change breadcrumbs logic to avoid these changes

PR #471, opened 2021-09-02
add breadcrumbs

VladyslavBondarenko:
List of breadcrumbs is formed from URL path.
If an URL part starts with `flutter-`, it's separated on 2 parts: `Workflow Editor` and the rest.
If an URL part starts with `yaml-`, it's separated on 2 parts: `.YAML` and the rest.
The last part is the article Title (to match the active menu item).


VladyslavBondarenko:
Note: breadcrumbs names as menu category names work as expected when the menu is nested max 2 layers. More layers will require a tiny change or introducing a function with recursion

assets/scss/modules/_breadcrumbs.scss
```diff
@@ -0,0 +1,39 @@
+#breadcrumbs {
+    display: flex;
+    align-items: center;
+    gap: 15px;
+    padding: 15px;
+    border-radius: 4px;
+    background-color: $blue-xl;
+    color: $grey-text-light;
```
mart-nc, line 8: Grey looks quite dirty on slightly blue background, use `#97a5c5`

mart-nc (approved):

helinanever (approved):
LGTM

PR #488, opened 2021-09-10
env variable groups

K-Codemagic:
Workflow samples changes - added env variable groups


helinanever:
I don't think it's enough to just replace the variables with the group name. Somewhere in the doc there should also be instructions to add those variables in the app settings or team settings with the group name and then import them into the yaml.

PR #531, opened 2021-09-23
Fix: role for Firebase distribution with service account

artemii-yanushevskyi:
https://codemagicio.slack.com/archives/CEKE2KZ37/p1632399999082700


helinanever (approved):
LGTM

PR #588, opened 2021-11-10
Fix: path to dSYM archive

artemii-yanushevskyi:
The current path to dSYM archives, `~/build.xcarchive` doesn't usually exist. Instead, we should use the usual paths for different Xcode versions


helinanever (approved):
LGTM

PR #660, opened 2021-12-29
Add aliased mapping under reuse sections

rudrankriyam:
Documentation for defining reusable sections under `definitions` and expanding them elsewhere using aliased mapping


nevercode-kevin (approved):
LGTM

PR #738, opened 2022-02-23
Update UNITY_HOME value. Use $UNITY_HOME when setting UNITY_BIN. Decode keystore to root of project.

nevercode-kevin:


remarkov (approved):
left minor suggestion in the comments

PR #743, opened 2022-03-03
Remove Unity section

nevercode-kevin:
Unity is now available on build machines so it's not necessary to have a dedicated host to get started building Unity projects now.


helinanever (approved):
LGTM

PR #788, opened 2022-04-26
Add correct build targets for Unity builds

nevercode-kevin:
When using addressables in Unity you should use the -buildTarget flag.


icarusdust (approved):
LGTM

PR #789, opened 2022-05-02
Update iOS Simulator to 15.4

rudrankriyam:
Small developer experience improvement while copying the command from the docs so that developers do not have to update 14.5 to 15.4 as `latest`, `edge` has iOS 15.4 simulators.


nevercode-kevin (approved):
LGTM

PR #793, opened 2022-05-06
Accesssing multiple private repos

martinjeret:
from thread: https://codemagicio.slack.com/archives/CEKE2KZ37/p1651831995847429?thread_ts=1651829263.857249&cid=CEKE2KZ37

### DIFF
--- a/content/configuration/access-private-git-submodules.md
+++ b/content/configuration/access-private-git-submodules.md
@@ -35,3 +35,18 @@ ssh-add /tmp/ssh_key
 ```
 
 But if you added a repository with an SSH key and want to use a different key to fetch dependencies, it's not possible to do in Workflow editor. Scripts are executed in independent shells, so the key explicitly added in a post-clone script will be lost as soon as the script finishes. The best thing to do in such a case would be to use the same key for both your repository and your private dependency. You may need to add the key to your account, not to the specific repository.
+
+One option if you have mulitple private repositories that cannot share the same key is using https links instead of ssh and replacing them during the build. An idea how you can achieve that:
+
+replace ssh links with the https links in pubspec.yaml with sed
+
+```bash
+      - &get_packages
+        name: Get packages
+        script: |
+          SSH="git@github.com:org/repo.git"
+          HTTPS="https://github.com/org/repo.git"
+          # using hash as delimiter for sed
+          sed -i '' 's#'"$SSH"'#'"$HTTPS"'#' pubspec.yaml
+          flutter packages get
+```


orestesgaolin:
As it's not possible to use multiple SSH keys, I had to take another route. Before getting the packages I just replace ssh URLs with https in the pubspec.yaml. Codemagic is able to access them if they are in the same org apparently.

```
      - &get_packages
        name: Get packages
        script: |
          SSH="git@github.com:org/repo.git"
          HTTPS="https://github.com/org/repo.git"
          # using hash as delimiter for sed
          sed -i '' 's#'"$SSH"'#'"$HTTPS"'#' pubspec.yaml
          flutter packages get
```


A solution which we use in our GitHub Actions is to use [this action](https://github.com/webfactory/ssh-agent) with keys per each repository like this:

```
      - name: Set SSH key for private repositories
        uses: webfactory/ssh-agent@v0.5.4
        with:
          ssh-private-key: |
            ${{ secrets.APP_GITHUB_KEY }}
            ${{ secrets.APP_GITHUB_KEY_1 }}
```

The keys have to be added as deploy keys to respective repositories. Moreover the `-C` argument while generating the keys [needs to include the repository url](https://github.com/webfactory/ssh-agent#exported-variables) e.g.:

```
ssh-keygen ... -C "git@github.com:owner/repo.git"
```

The action matches the key for url, so that it's possible to use multiple deploy keys at the same time.

orestesgaolin (approved):

mikhail-tokarev:
> But if you added a repository with an SSH key and want to use a different key to fetch dependencies, it's not possible to do in Workflow editor. Scripts are executed in independent shells, so the key explicitly added in a post-clone script will be lost as soon as the script finishes

this part looks obsolete to me. You can add different SSH keys as environment variables with `SSH_KEY` suffix and they will be automatically added to SSH agent. I'll close this PR for now

orestesgaolin:
>  You can add different SSH keys as environment variables with SSH_KEY suffix and they will be automatically added to SSH agent.

Does it work if you have multiple private repositories added via ssh urls? I had 2 repositories like that and only one SSH_KEY can be used at a time, whereas deploy keys are different for each repository. Did something change about it? In the action I linked they have to do a workaround with the key comment:

> When using Github deploy keys, GitHub servers will accept the first known key. But since deploy keys are scoped to a single repository, this might not be the key needed to access a particular repository. Thus, you will get the error message fatal: (...).
> To support picking the right key in this use case, this action **scans key comments and will set up extra Git and SSH configuration to make things work**.

PR #807, opened 2022-05-29
Add instance types to machine spec pages

aperekalin:
A small change to improve navigating the docs, as getting to the https://docs.codemagic.io/yaml/yaml-getting-started/#instance-type page may be quite hard.


helinanever:
I see some value in adding the instance types to the machine-type file but I'm not sure about adding them to base image specs, as the instance types are relevant for yaml users only.

helinanever:
@aperekalin The CE team is planning a full rewrite of the documentation, including this page, so I would rather close this PR for now without proceeding with the changes.

PR #823, opened 2022-06-13
Add access policy snippet and recommendation to untick raw message delivery

VeArnold:
Raw message delivery results in some circumstances in unparseable messages, thus it is recommended to not use it.

In addition, a snippet was added for configuring access policy, in case the default configuration provided by AWS is not sufficient.


helinanever (approved):
LGTM

PR #856, opened 2022-06-30
Add border-bottom underline to the active tab

todah-zg:
Just an idea to make the active tab more prominently visible by using an underline in our brand color (magenta)


mart-nc:
Will improve visibility a bit differently in a future pull request.

PR #857, opened 2022-07-01
Update build history retention policy

helinanever:
Updated info about build history and artifacts retention to reflect upcoming changes.


nevercode-kevin (approved):
LGTM

Issue #803, opened 2022-05-25
Add warning about special character in release notes which break publishing release notes to AppStore Connect


Ycoi:
See related slack thread: https://codemagicio.slack.com/archives/CELDFDE66/p1653422291989739

Some special characters like `<` and `>` in the release notes file break publishing release notes to AppStore Connect (AppStore Connect API will return 409).  

Would be good to have warning about this in the Codemagic documentation https://docs.codemagic.io/flutter-publishing/publish-release-notes/ Especially since Apple Documentation doesn't mention this explicitly.

PR #860, opened 2022-07-05
Add a warning

todah-zg:
Add a warning about unsupported symbols in release notes (from Issue #803)

### DIFF
--- a/content/yaml-publishing/publish-release-notes.md
+++ b/content/yaml-publishing/publish-release-notes.md
@@ -68,5 +68,5 @@ For App Store Connect supported languages and codes are listed [here](https://de
     
     
 {{<notebox>}}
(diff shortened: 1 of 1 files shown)


content/yaml-publishing/publish-release-notes.md
```diff
@@ -65,3 +65,8 @@ For App Store Connect supported languages and codes are listed [here](https://de
         }
     ]
     ```
+    
+    
+{{<notebox>}}
+**Warning**: Apple does not support `<` and `>` symbols in `release_notes` file. Uploading a file with such symbols will cause the AppStore Connect API to return `409` error and description that text contains invalid characters.
```
nevercode-kevin, line 71:
```suggestion
**Warning**: Apple does not support `<` and `>` symbols in `release_notes` file. Uploading a file with such symbols will cause the App Store Connect API to return `409` error and description that text contains invalid characters.
```

nevercode-kevin (changes requested):

nevercode-kevin (approved):
LGTM

PR #866, opened 2022-07-05
Add comments to markdown shortcode

todah-zg:
Markdown shortcode does not properly create paragraphs if it wraps only a single paragraph of text. Adding another paragraph, even if it's an empty one is a passable workaround.


nevercode-kevin (approved):
LGTM

PR #867, opened 2022-07-05
Split the guide into partials...

todah-zg:
... that can later be reused in other guides. Remove HTML in favor of Markdown wherever possible.

### DIFF
--- a/content/partials/build-versioning-ios.md
+++ b/content/partials/build-versioning-ios.md
@@ -1,7 +1,7 @@
 ---
 ---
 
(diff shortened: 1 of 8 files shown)


content/partials/build-versioning-ios.md
```diff
@@ -0,0 +1,30 @@
+---
+---
+
+In addition to the App Store credentials configured earlier, in order to get the latest build number from App Store or TestFlight, you will also need the **Application Apple ID**. This is an automatically generated ID assigned to your app and it can be found under **General > App Information > Apple ID** under your application in App Store Connect.
```
nevercode-kevin, line 4: Partials should be standalone so they can be reused anywhere. Therefore, I would re-word this slightly to remove the reference to things that were configured earlier.

nevercode-kevin (changes requested):

nevercode-kevin (approved):
LGTM

PR #877, opened 2022-07-18
Update build-versioning.md

himesh-cm:
Better clarification regarding the build command for flutter.

### DIFF
--- a/content/configuration/build-versioning.md
+++ b/content/configuration/build-versioning.md
@@ -47,7 +47,7 @@ For Flutter, you can easily increment your build number and build name using the
 --build-name=1.0.$PROJECT_BUILD_NUMBER --build-number=$PROJECT_BUILD_NUMBER
 ```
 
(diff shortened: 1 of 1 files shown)


content/configuration/build-versioning.md
```diff
@@ -47,7 +47,7 @@ For Flutter, you can easily increment your build number and build name using the
 --build-name=1.0.$PROJECT_BUILD_NUMBER --build-number=$PROJECT_BUILD_NUMBER
 ```
 
-Note I've added `--build-name` as an example, if you've committed this to version control in `pubspec.yaml` version field, then don't pass in this argument to builds.
+Note that I’ve added `--build-name` & `--build-number` as an example. Don’t pass both arguments to the build command if you’ve committed this to version control in the `pubspec.yaml` version field.
```
icarusdust, line 50: What about something like this?

Note that`--build-name` & `--build-number` added as an example above. If the build version is manually incremented in `pubspec.yaml`, then these arguments are not needed in the build command.

icarusdust (changes requested):

icarusdust (approved):
LGTM

PR #901, opened 2022-08-12
Added build script for Windows to Unity docs

aperekalin:
Previously the build script only had items for Android and iOS. I think it's worth adding workflows for Windows there as well. Maybe there are other parts of the Codemagic Unity docs that we need to pay attention to, but the build script is obviously missing. Borrowed one from an upcoming article by Mina.

### DIFF
--- a/content/yaml-quick-start/building-a-unity-app.md
+++ b/content/yaml-quick-start/building-a-unity-app.md
@@ -189,6 +189,20 @@ public static class BuildScript
         Debug.Log("Built iOS");
     }
 
+    [MenuItem("Build/Build Windows")]
+    public static void BuildWindows()
+    {
+        BuildPlayerOptions buildPlayerOptions = new BuildPlayerOptions();
+        buildPlayerOptions.locationPathName = "win/" + Application.productName + ".exe";
+        buildPlayerOptions.target = BuildTarget.StandaloneWindows;
+        buildPlayerOptions.options = BuildOptions.None;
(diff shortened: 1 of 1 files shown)


content/yaml-quick-start/building-a-unity-app.md
```diff
@@ -189,6 +189,20 @@ public static class BuildScript
...
+        buildPlayerOptions.locationPathName = "win/" + Application.productName + ".exe";
+        buildPlayerOptions.target = BuildTarget.StandaloneWindows;
+        buildPlayerOptions.options = BuildOptions.None;
+        buildPlayerOptions.scenes = GetScenes();
+
+        Debug.Log("Building Windows");
+        BuildPipeline.BuildPlayer(buildPlayerOptions);
+        Debug.Log("Built Windows");
+    }
+
```
nevercode-kevin, line 205: This page is a quick start for building 'Building Unity **mobile** apps' for Android and iOS. 

Not sure how adding the script for building a Windows desktop fits in here.

aperekalin: We don't have any docs for Unity & Windows. Even for the Steam publishing, where Windows is the main platform. If not here, then we need some other place for that.

nevercode-kevin: Let's put it here for now. We can possibly come up with a better solution at a later date.

nevercode-kevin (approved):
LGTM

PR #930, opened 2022-09-02
Add missing dots to Google Play Android page

KeisukeYamashita:
## WHAT

I realized that there are some comments which is missing `.`. This PR adds it.


helinanever (approved):
LGTM, thanks for the contribution!

PR #967, opened 2022-09-20
Big batch of changes

todah-zg:


todah-zg:
This is a massive batch of changes. Many pages were moved around so it's best to review and inspect locally, compare with the current online version and approve only when we are ready to change everything.
Comments on particular pages are welcome.

Work in progress:
TODO:
- some pages still left to improve (menu item 'Advanced Configuration > Various' and below)
- Quick start guides need a rewrite to make them a standalone A-Z guide for their respective platform
- WFE menu needs adjusting - I will rearrange the menu to closely mirror the new Yaml menu (better grouping and organization)

config.yaml
```diff
@@ -62,10 +68,41 @@ menu:
...
+        - identifier: knowledge-codemagic
+          name: Codemagic
+          weight: 1
+          parent: knowledge-base
+        - identifier: knowledge-firebase
+          name: Firebase
+          weight: 3
+          parent: knowledge-base
+        - identifier: knowledge-publishing
+          name: Publishing
```
helinanever, line 99: I'm not entirely convinced it's a good idea to duplicate subcategory names under Configuration and Advanced configuration. Perhaps there's a better way to make the distinction between which publishing articles should be in the two parent categories. One idea for Configuration > Publishing would be to keep this for built-in features and/or stores, and everything else in Advanced configuration. Or make the subcategory in Advanced configuration about uploading artifacts and change the name accordingly.

content/knowledge-firebase/firebase-crashlytics-dsym-uploading.md
```diff
@@ -0,0 +1,89 @@
+---
+description: How to upload dsym artifacts to Firebase Crashlytics
+title: Firebase Crashlytics dsym uploading
```
helinanever, line 3: "dsym" here but "dSYM" in the Sentry file name

helinanever (commented):
Left some in-line comments.

Re build machine specs:
Overall I like the tab structure but I would recommend making changes to these files in a separate PR as these changes would have to be aligned with the automated reports coming from the builder when we update base images. Additionally, it would be nice to also have instance types mapped to the hardware in specs.

helinanever:
Also, either before merging the docs or after deployment, we should run a crawler to help find possible broken links as the changes to docs structure are substantial in this PR.

config.yaml
todah-zg: So far, the 'Publishin' section in 'Advanced Configuration' only has 2 items so I would suggest we simply move them to 'Advanced configuration > Various'

helinanever: Sounds good.

PR #1021, opened 2022-10-06
Docs: Using team or account App Store Connect integration from Yaml workflows

priitlatt:


content/partials/code-signing-ios-configure-environment-vars.md
```diff
@@ -3,12 +3,58 @@
 
 #### Configuring environment variables
 
+Provisioning profiles and code signing certificates are obtained from Apple Developer portal with command [`app-store-connect fetch-signing-files`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#fetch-signing-files). App Store Connect API key information can be passed to it via environment variables [`APP_STORE_CONNECT_KEY_IDENTIFIER`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#--key-idkey_identifier), [`APP_STORE_CONNECT_ISSUER_ID`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#--issuer-idissuer_id), [`APP_STORE_CONNECT_PRIVATE_KEY`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#--private-keyprivate_key).
```
helinanever, line 6:
```suggestion
Provisioning profiles and code signing certificates are obtained from Apple Developer portal with the command [`app-store-connect fetch-signing-files`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#fetch-signing-files). App Store Connect API key information can be passed to it via environment variables [`APP_STORE_CONNECT_KEY_IDENTIFIER`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#--key-idkey_identifier), [`APP_STORE_CONNECT_ISSUER_ID`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#--issuer-idissuer_id), [`APP_STORE_CONNECT_PRIVATE_KEY`](https://github.com/codemagic-ci-cd/cli-tools/blob/master/docs/app-store-connect/fetch-signing-files.md#--private-keyprivate_key).
```

content/partials/code-signing-personal-ios.md
```diff
@@ -1,25 +1,29 @@
...
+When automatic code signing is used, then most up-to-date signing files are obtained directly from Apple during the build time. This requires that Codemagic has access to your Apple Developer portal account, which is achieved by using App Store Connect API key.
 
-To code sign the app, add the following commands in the [`scripts`](../getting-started/yaml#scripts) section of the configuration file, after all the dependencies are installed, right before the build commands. 
+#### Creating the App Store Connect API key
+
+{{< include "/partials/app-store-connect-api-key.md" >}}
+
+{{< include "/partials/code-signing-ios-obtain-certificate.md" >}}
+
+{{< include "/partials/code-signing-ios-configure-environment-vars.md" >}}
```
helinanever, line 24: Including this here makes the whole automatic code signing part extremely difficult to follow for me, mainly due to sections being nested inside sections. It remains unclear why the Apple Developer Portal integration section is part of "Configuring environment variables". I believe the only reason for this is to cover adding the private key as an env var? And why is that even needed if we're using the integration to authenticate?

helinanever (changes requested):
The App Store Connect publishing section looks good to me but there is some confusion in the code signing part.

content/partials/code-signing-personal-ios.md
priitlatt: Certificate private key has nothing to do with ASC/ADP authentication and is a standalone thing that is needed whenever working with certificates that are downloaded from Dev portal as for code signing one need both the secret part (certificate private key) and the certificate itself. Apple gives you only the latter while the former needs comes from user.
In case of manual signing certs are stored as PKCS#12 containers which have both certificate and private key within one single file.

helinanever (approved):
Approving, as the structure changes to code signing docs are out of the scope of this PR.

PR #1026, opened 2022-10-07
Update building-a-flutter-app.md

K-Codemagic:
Flutter build ipa


todah-zg (approved):
Thanks for catching this

PR #1031, opened 2022-10-07
Fix increment of BUILD_NUMBER

joelpaulkoch:
BUILD_NUMBER was not incremented, but the latest one was used which results in an error.


todah-zg (approved):
Thank you for noticing this and correcting

PR #1107, opened 2022-10-29
Updating testing.md fixing wrong indentation for "test_report"

Flowdawan:
**Problem:** Indentation of the "test_report" is not correct and leads to a "Configuration file error".

**Solution:** Correct the indentation of the test_report:
From:
 - name: test
      script: ./gradlew test
 test_report: app/build/test-results/**/*.xml
to: 
 - name: test
      script: ./gradlew test
      test_report: app/build/test-results/**/*.xml

**More details:**
The indentation in the yaml file of the "_test_report_" is not correct and therefore leads to a visible "_Configuration file error_" in the yaml overview on the codemagic page itself.

I could only test this with a native Android app, but it's possible that the indentation is also wrong for the other test cases in the documentation (flutter, ..). For "Native macOS" the indentation in the documentation should be correct and is as I suggested.

Exact error message in the codemagic.yaml overview (on codemagic.io):

"
```
Configuration file error:
While parsing a block collection
  in "<byte string>", line 12, column 7:
          - name: set android sdk location
          ^
expected <block end>, but found '?'
  in "<byte string>", line 17, column 7:
          test_report: app/build/test-resu ... 
          ^
```
"

Thanks!

### DIFF
--- a/content/yaml-quick-start/building-a-native-android-app.md
+++ b/content/yaml-quick-start/building-a-native-android-app.md
@@ -54,7 +54,7 @@ scripts:
       # gradlew assembleRelease # -> to create the .apk
 
 artifacts:
-  - app/build/outputs/**/*.aab
+  - android/app/build/outputs/**/*.aab
 {{< /highlight >}}
 
 ## Build versioning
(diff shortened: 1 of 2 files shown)


todah-zg (approved):
Thank you for finding and correcting this

PR #1120, opened 2022-11-01
Use checkout@v3

VeArnold:
Currently deployment builds have the warning:
```
Node.js 12 actions are deprecated. For more information see: [GitHub Actions: All Actions will begin running on Node16 instead of Node12 | GitHub Changelog](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/). Please update the following actions to use Node.js 16: actions/checkout, actions/checkout
```

Updating checkout from v2 to v3 solves the issue, using node 16 by default. 
https://github.com/actions/checkout#whats-new


mikhail-tokarev (approved):
🚀

PR #1123, opened 2022-11-02
Small error corrections

todah-zg:


content/flutter-configuration/built-in-variables.md
```diff
@@ -1,6 +1,7 @@
 ---
-description: Useful variables exported during builds
-title: Built-in environment variables
+description: Useful variables exported during builds using Flutter Workflow Editor
```
aperekalin, line 2: Do we capitalize all first letters in Flutter Workflow Editor or do we capitalize on the F in Flutter?

content/flutter-configuration/built-in-variables.md
```diff
@@ -1,6 +1,7 @@
...
+title: Built-in environment variables using WFE
```
aperekalin, line 3: I'd rather go for Flutter Workflow Editor instead of WFE (if the length of the title is still below 60 symbols), because Flutter is a relevant keyword, and WFE is not. Similarly in other places.

content/yaml-publishing/aws.md
```diff
@@ -1,5 +1,6 @@
 ---
-title: Amazon S3
+title: Amazon S3 publishing using yaml
```
aperekalin, line 2: maybe _Amazon S3 publishing using codemagic.yaml_ ?

content/flutter-configuration/built-in-variables.md
todah-zg: Good point. I don't know for sure. Statistically, we use 'Flutter workflow editor' more often but it could be that I introduced that statistic with recent docs rewrite. Our main page (https://docs.codemagic.io/) has 'Flutter workflow editor" above but then also "Flutter Workflow Editor' immediately below if you scroll just a bit.

I am ok with both so please choose one :-D

aperekalin: Looked at the blog and it's the same. Let it be `Flutter workflow editor` similarly to how it's `Codemagic` rather than `CodeMagic` :)

aperekalin (approved):
LGTM

PR #1161, opened 2022-11-16
Sync docs with behaviour

VeArnold:
The actual behaviour differs from what is described in the documentation, as the global package is always ran when dart code metrics is enabled.


helinanever (approved):
LGTM

PR #1226, opened 2022-12-14
Update iOS Code Signing to mention App Extensions for Profiles

nevercode-rudrank:
While it is mentioned for the workflow editor that you are required to upload provisioning profiles for each app extension, it is helpful to mention it for yaml too.


nevercode-rudrank:
Oops looks like I forgot to remove it from drafts.

nevercode-kevin (approved):
LGTM

PR #1332, opened 2023-01-19
Enterprise plan with dedicated hosts

nevercode-kevin:
Add details about dedicated host pricing when using the Enterprise plan.


todah-zg (approved):
LGTM

PR #1383, opened 2023-02-06
Firebase publishing now uses service account instead of token as default

todah-zg:


content/yaml-publishing/firebase-app-distribution.md
```diff
@@ -180,28 +151,25 @@ buildTypes {
...
-<br>{{<markdown>}}
-9. Decode application credentials for Firebase authorization:{{</markdown>}}
+2. Decode application credentials for Firebase authorization:
 
 {{< highlight yaml "style=paraiso-dark">}}
 scripts:
   -name: Decode Google credentials
    script: | 
-     echo $GOOGLE_APP_CREDENTIALS > $CM_BUILD_DIR/your/file/path.json
+     echo $FIREBASE_SERVICE_ACCOUNT > $GOOGLE_APPLICATION_CREDENTIALS
```
nevercode-kevin, line 166: Might want to consider changing the variable name to something that reflects it's a file path e.g. `GOOGLE_APPLICATION_CREDENTIALS_PATH` in case people start adding the JSON value to it by mistake?

nevercode-kevin: Keep it like that then to be consistent.

PR #1408, opened 2023-02-15
fix typo on android app signing page

UserComponent:
caught it while browsing docs 👍🏻


helinanever (approved):
LGTM, thanks a lot for fixing this!

PR #1469, opened 2023-03-13
Update Machine Type

nevercode-rudrank:
As the Intel Mac mini is now deprecated, update the machine type to consist only of Mac Pro and M1 Mac mini


nevercode-kevin (approved):
LGTM

PR #1496, opened 2023-03-20
Fix: increase default clone depth for destination

artemii-yanushevskyi:
GitHub PR webhook payload specifies latest destination commit, instead of the latest commit that is also available in source branch

This caused a few SWWs since the deployment of clone depth PRs

This seem to happen when PR destination branch had new pushes that are not yet merged in the source branch


helinanever (approved):
LGTM

PR #1544, opened 2023-04-05
Add information about setting Flutter versions along with FVM

VeArnold:
This PR updates both the Flutter WFE and yaml to add short paragraphs about Flutter version management with Codemagic, along with FVM itself.


helinanever (approved):
Very nice, LGTM

PR #1548, opened 2023-04-06
update billing info

martinjeret:
we got email about user asking how to update taxID and turns out it's not available on personal accounts. Updating docs to reflect that.


helinanever (approved):
LGTM

PR #1560, opened 2023-04-12
Update billing.md

martinjeret:
We don't charge for users anymore, but this entry was still in docs. probably missed when we announced new pricing.


helinanever (approved):
LGTM

PR #1607, opened 2023-05-01
Add example on how to set the `instance_type` via the API

nilsreichardt:
I wondered how to change the `instance_type` via the API. Unfortunately, I couldn't find any information about this in the documentation. Therefore, I tried `instance_type` first. This hasn't worked. Then I tried `instanceType`. This worked :)

Therefore, I thought, it would be maybe helpful to mention this in the documentation ["Pass custom build parameters" section](https://docs.codemagic.io/rest-api/builds/#pass-custom-build-parameters).


icarusdust (approved):
LGTM

PR #1762, opened 2023-06-13
Update email-and-slack-notifications.md

rafeysup93:
Added .exe support for Artifacts and a note that emails will only be sent on succesful builds with the artifacts.


icarusdust (approved):
lgtm

PR #1792, opened 2023-06-19
A required agreement is missing or has expired

dtrdic:
Updated common iOS issue page about the "A required agreement is missing or has expired" error


icarusdust (changes requested):

icarusdust (approved):
lgtm

PR #1807, opened 2023-06-23
reinstate Flutter iOS auto codesigning video embed

maks:
For some reason this video embed was lost at some point, so re-instating it, as its still relevant to the topic.


icarusdust (approved):
lgtm

PR #1814, opened 2023-06-26
add env var CM_WORKFLOW_NAME

fran-tirapu:
With this [PR](https://github.com/codemagic-dev/builder_worker/pull/103), we add a new environment variable available for all builds, so we should add it to our docs.


helinanever (approved):
LGTM

PR #1816, opened 2023-06-27
add utm tags to codelab links

maks:
This adds utm tags that were missed out in previous addition of the links to the codelab.


icarusdust (approved):
lgtm

PR #1818, opened 2023-06-27
separated instructions for wfe and yaml for uploading dSYM file to Firebase

dtrdic:
added a separate, much more visible step for uploading dSYM file to Firebase, since we have a lot of queries on how to do it with wfe


icarusdust:
Why do we want to duplicate the script ? I mean it says add the script to post-publishing step in the WFE and in yaml.

dtrdic:
> Why do we want to duplicate the script ? I mean it says add the script to post-publishing step in the WFE and in yaml.

many users don't notice that sentence, they just see that is about yaml file, and stop reading further. This was the reason I decided to make it more obvious. I can close this pr if you don't agree

icarusdust:
Let's see how it looks then

icarusdust (approved):
lgtm

PR #1821, opened 2023-06-27
delete account instructions

dtrdic:


icarusdust:
Is adding apps to CM page a correct place for this info ?

dtrdic:
> Is adding apps to CM page a correct place for this info ?

yes I agree, actually not a good spot for that information.. Can I rather move it [here](https://docs.codemagic.io/getting-started/teams/#deleting-the-team)? below deleting team heading

icarusdust:
Maybe let's add it to [FAQ's page ](https://docs.codemagic.io/getting-started/faq/)? Deleting a team should be different than deleting account

icarusdust:
Also, if they need us to delete any chat threads related to their account, they will need to reach out to us.

icarusdust (approved):
LGTM

PR #1858, opened 2023-07-07
Update common-android-issues.md

rafeysup93:
Added note to use Xcode version 14.2 for Mac Pro machines.


icarusdust:
It is not necessarily 14.2, it is just not above 14.2

icarusdust:
Maybe use that note under the solution section and say that if cannot upgrade gradle and AGP versions, then you can use Mac Pro machines, and then give this note.

icarusdust (approved):

PR #1950, opened 2023-08-08
1949 api information about yaml

kulakarla:
Added this note to 2 yaml api pages.

{{}} **Note:** Using our API you cannot fetch information about workflows when configuring with **codemagic.yaml**, as we do not save YAML configurations to our database. Values regarding workflows will be null or false. {{}}


icarusdust (approved):
LGTM

PR #2005, opened 2023-08-29
Improvement: deprecate Mac mini

artemii-yanushevskyi:
Remove Mac mini machine specs following deprecation.


fran-tirapu (changes requested):
May be we should include a deprecation warning in our docs?

fran-tirapu (approved):

PR #2151, opened 2023-10-25
Deprecation: Standard Linux instance type

artemii-yanushevskyi:
Documentation changes to reflect build machine naming changes and the deprecation of Standard Linux instance type: https://github.com/orgs/codemagic-ci-cd/discussions/2063


helinanever (approved):
LGTM

PR #2175, opened 2023-11-07
Update building-an-ionic-app.md

bulletshot60:
The documentation page has a wrong command.  Hopefully will save another dev a few minutes of time.


bulletshot60:
The command was correct, my build.gradle was outdated.

PR #2210, opened 2023-11-21
Add new `pull_request_labeled` triggering event

mohammedbabelly20:
We now have a new triggering event called `pull_request_labeled` which triggers builds when a new label is added to a PR.

### DIFF
--- a/content/yaml-running-builds/starting-builds-automatically.md
+++ b/content/yaml-running-builds/starting-builds-automatically.md
@@ -117,10 +117,10 @@ triggering:
 {{< /highlight >}}
 
 {{<notebox>}}
-**Note:** Above pattern is set for `main` branch but you can set similar patterns for any branch depending on your workflow.
+**Note:** The Above pattern is set for the `main` branch but you can set similar patterns for any branch depending on your workflow.
 {{</notebox>}}
 
-**Example 3**. Trigger a new build when adding new labels to your pull requests.
+**Example 3**. Trigger a build when adding a label to your pull request.
 
 {{< highlight yaml "style=paraiso-dark">}}
 triggering:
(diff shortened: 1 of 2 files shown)


content/yaml-running-builds/starting-builds-automatically.md
```diff
@@ -321,20 +335,20 @@ workflows:
       condition: not event.pull_request.draft
 {{< /highlight >}}
 
-Example 2. This build will *not* continue if the pull request branch has a specific label name, "codemagicTest" in the following sample:
+**Example 2**. This build will *not* continue if the new label added to a pull request is not a specific label name, "codemagicTest" in the following sample:
```
helinanever, line 338: This explanation doesn't seem to go with the example as you have a double negative here ("will not continue" and "is not a specific label name").
```suggestion
**Example 2**. This build is triggered on adding a label to a pull request but will continue only if the label added was anything else than "codemagicTest":
```

content/yaml-running-builds/starting-builds-automatically.md
```diff
@@ -321,20 +335,20 @@ workflows:
...
       events:
-        - pull_request
+        - pull_request_labeled
     when:
-      condition: not event.pull_request.labels[0].name == "codemagicTest"
+      condition: not event.pull_request.label.name == "codemagicTest"
 {{< /highlight >}}
 
-Example 3. You can also combine triggering conditions, just make sure that each condition is wrapped with brackets:
+**Example 3**. You can also combine triggering conditions, just make sure that each condition is wrapped with brackets:
```
helinanever, line 351:
```suggestion
**Example 3**. You can also combine triggering conditions, just make sure that each condition is wrapped in brackets:
```

helinanever (changes requested):

PR #2211, opened 2023-11-21
Add new `pull_request_labeled` triggering event

mohammedbabelly20:
Mention that the new `pull_request_labeled` event is only for GitHub repos.


mohammedbabelly20:
Do you think this is enough? @helinanever

helinanever (approved):
LGTM

PR #2229, opened 2023-11-29
Add Legacy Flag to Private Key Export

hschaeufler:
When the command `openssl pkcs12 -in IOS_DISTRIBUTION.p12 -nodes -nocerts | openssl rsa -out ios_distribution_private_key` is used to extract the private key it comes to the following Error.

```
Error outputting keys and certificates
C0DE35E001000000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:341:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
Could not read private key from <stdin>
```

As stated in [Issue 5059](https://github.com/Kong/insomnia/issues/5059), you can add the legacy Flag to fix this.

### DIFF
--- a/content/partials/code-signing-ios-obtain-certificate.md
+++ b/content/partials/code-signing-ios-obtain-certificate.md
@@ -36,7 +36,7 @@ This new private key will be used to create a new iOS Distribution certificate i
 7. Use the following `openssl` command to export the private key:
 
 {{< highlight Shell "style=rrt">}}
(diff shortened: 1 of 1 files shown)


icarusdust:
Hi @hschaeufler, it looks more like an issue with the private key that your `.p12` certificate holds. I quickly tested it with my iOS distribution certificate where a private key is in place, and I did not get any warnings or errors with the command without the `--legacy` flag. 
Screenshot 2024-01-11 at 09 45 10

icarusdust:
I will close the PR for now @hschaeufler Let me know if you want me to review it again, please.

PR #2293, opened 2024-01-03
Update commands to list and create certificates

priitlatt:
Some Codemagic CLI tools were renamed in https://github.com/codemagic-ci-cd/cli-tools/pull/386. This PR ensures that deprecated names are not referenced in current docs.


helinanever (approved):
LGTM

PR #2302, opened 2024-01-10
Update environment-variables.md

himesh-cm:
CM_PREVIOUS_COMMIT supports only within the same branch. Documentation needs to clarify that it is branch-specific.
i have tested the same.


kevin-suhajda (approved):
LGTM

PR #2319, opened 2024-01-18
Update common-android-issues.md

himesh-cm:
removed the line - 
If you can’t upgrade the Gradle and AGP versions, then you can use Mac Pro machines instead.


kevin-suhajda (approved):
LGTM

PR #2343, opened 2024-01-31
Update android-code-signing.md

himesh-cm:
Clarified that this change needs to be done on app-level build.gradle file


kevin-suhajda (approved):
LGTM

PR #2397, opened 2024-02-29
Did not find xcodeproj

dtrdic:
Solution for _Did not find xcodeproj from /Users/builder/clone/ios_ error

image


icarusdust (approved):
LGTM

PR #2438, opened 2024-03-26
Update code-signing-macos.md

himesh-cm:
Notarizing macs app using notary tool


icarusdust:
If it is not ready @himesh-cm, you can convert it to draft

PR #2442, opened 2024-03-27
update firewall requirements

mikhail-tokarev:
- add Las Vegas
- improve description and formatting (now it's possible to copy all addresses to clipboard)
- TBD new Atlanta

### DIFF
--- a/content/partials/firewall.md
+++ b/content/partials/firewall.md
@@ -5,7 +5,7 @@ weight: 7
 aliases:
 ---
 
-If your repositories are within an internal network/behind a firewall, which is usually the case with enterprises, the network configuration of these repositories must be configured for external access.
+If your repositories are within an internal network / behind a firewall, which is usually the case with enterprises, the network configuration of these repositories must be configured for external access.
 
 Codemagic requires direct access to the repositories for the following use cases:
 - For retrieving the repository information such as the branches and the commits.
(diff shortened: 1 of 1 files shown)


content/partials/firewall.md
```diff
@@ -5,10 +5,18 @@ weight: 7
 aliases:
 ---
 
-To allow Codemagic access a private repository, the following IP addresses need to be whitelisted:
+If your repositories are within an internal network/behind a firewall, which is usually the case with enterprises, the network configuration of these repositories must be configured for direct access.
```
mikhail-tokarev, line 8:
```suggestion
If your repositories are within an internal network/behind a firewall, which is usually the case with enterprises, the network configuration of these repositories must be configured for external access.
```

content/partials/firewall.md
```diff
@@ -5,10 +5,18 @@ weight: 7
 aliases:
 ---
 
-To allow Codemagic access a private repository, the following IP addresses need to be whitelisted:
+If your repositories are within an internal network/behind a firewall, which is usually the case with enterprises, the network configuration of these repositories must be configured for external access.
```
helinanever, line 8:
```suggestion
If your repositories are within an internal network / behind a firewall, which is usually the case with enterprises, the network configuration of these repositories must be configured for external access.
```

helinanever (approved):
Added one suggestion but otherwise LGTM

PR #2468, opened 2024-04-12
added github checks and build dashboards page to WFE Various section

dtrdic:
Some generic docs are available in the left menu only for yaml but not for Flutter Workflow Editor. For example, Build dashboards and GitHub checks.
Added this info on the Various section in WFE
image

image


icarusdust (approved):
lgtm

PR #2500, opened 2024-04-30
using non-default ruby only for macOS

mohammedbabelly20:
We only support installing another version of Ruby on macOS at the moment.


helinanever (approved):
LGTM

PR #2513, opened 2024-05-07
Update yaml-getting-started.md

martinjeret:
people are confused about what source means and where to use it.


icarusdust (approved):
lgtm

PR #2518, opened 2024-05-08
fix Milwaukee IP address

mikhail-tokarev:
`66.185.17.82/32` is correct address, but we should use subnet address. 
So users no need to update firewall rules every time we add more hosts.

### DIFF
--- a/content/partials/firewall.md
+++ b/content/partials/firewall.md
@@ -12,12 +12,13 @@ Codemagic requires direct access to the repositories for the following use cases
 - For cloning the repository to the build machine during the build.
 
 The following IP addresses must be allowed through the firewall (whitelisted):
+
 {{< highlight Shell "style=rrt">}}
+66.185.17.80/29
 199.7.162.128/29
 207.254.42.240/29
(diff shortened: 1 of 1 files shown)


content/partials/firewall.md
```diff
@@ -12,12 +12,13 @@ Codemagic requires direct access to the repositories for the following use cases
...
 
 The following IP addresses must be allowed through the firewall (whitelisted):
+
 {{< highlight Shell "style=rrt">}}
+66.185.17.80/29
 199.7.162.128/29
 207.254.42.240/29
 34.74.234.56/32
 35.185.76.207/32
-66.185.17.82/32
```
dtrdic, line 20: you are saying that 66.185.17.82/32 is the correct address, how come we are removing it from the docs then?

mikhail-tokarev: I didn't remove it, just moved as the first one

dtrdic: ah, my mistake..
Looks good to me now

dtrdic (approved):

PR #2525, opened 2024-05-14
Remove release log

helinanever:
We don't often update the release log anymore and there is no point in keeping a stale page in the docs.


helinanever:
@AstDaria I couldn't figure out how to reduce the spacing after the Software & Hardware section in the left navigation menu, I hope you can help with that.

AstDaria (approved):

PR #2548, opened 2024-05-23
Update yaml-validate.md to point correct link to the schema

bahag-chandrana:
The link for the schema in the documentation is wrong and points to an old schema. Updated the schema link to the latest one.


icarusdust (approved):
LGTM

PR #2585, opened 2024-06-07
Adding build inputs

icarusdust:


content/yaml-basic-configuration/yaml-getting-started.md
```diff
@@ -150,6 +155,10 @@ workflows:
         - echo 'Post-publish script'
 {{< /highlight >}}
 
+### Build inputs
+
+Build inputs are parameters that allow you to customize your build configurations right before starting a new build without adding them in **codemagic.yaml** manually. For example, build inputs can be used to determine whether to build the workflow for test or release purposes or which Xcode version to use etc. More information about how to configure build inputs and real time examples can be found (here)[https://docs.codemagic.io/knowledge-codemagic/build-inputs/].
```
helinanever, line 160:
```suggestion
Build inputs are parameters that allow you to customize your build configurations right before starting a new build without hardcoding them in **codemagic.yaml**. For example, build inputs can be used to determine whether to build the workflow for test or release purposes or which Xcode version to use, etc. More information about how to configure build inputs and examples can be found [here](https://docs.codemagic.io/knowledge-codemagic/build-inputs/]).
```

helinanever: Move the section after Instance type, so it follows the same order as in the template.

icarusdust: done

helinanever (approved):
LGTM

PR #2618, opened 2024-06-28
Remove TestFlight submission requirement for beta groups

helinanever:
We no longer require submitting builds to TestFlight beta review as a prerequisite to distributing to beta groups.


priitlatt (approved):
🚢

PR #2623, opened 2024-07-04
Add note about labels being supported for team apps only

helinanever:
The docs do not specify that labels work with team apps only. When using them with personal apps, labels are not displayed in the build overview or on the builds page.


helinanever:
Closing as this note is incorrect and labels work with personal apps too.

PR #2683, opened 2024-08-14
Update App Store Connect API URLs

priitlatt:
iOS code signing sections that describe App Store Connect API key generation process reference outdated App Store Connect links.


helinanever (approved):
LGTM

PR #2730, opened 2024-09-11
Remove references to M1

helinanever:
Now that we have removed M1 machines, we can remove outdated references to M1 instances.


icarusdust (approved):
LGTM

PR #2828, opened 2024-11-14
Fix the "Get Flutter packages" script

sbis04:
Update the Flutter packages script from `flutter packages pub get` to `flutter pub get`


mikhail-tokarev (approved):
LGTM

PR #2835, opened 2024-11-16
use dedicated pages for macOS machine specs

mikhail-tokarev:
1. use navigation bar to display available Xcode versions. I order them from newer to older versions.
2. replace page descriptions with paragraph text. The copy is open for discussion.

open question: I think we should move `Xcode versions` section before `Pre-installed tools`

Screenshot 2024-11-18 at 10 52 25


helinanever:
This isa good idea
>open question: I think we should move Xcode versions section before Pre-installed tools

content/specs-macos/xcode-15-1.md
```diff
@@ -1,19 +1,22 @@
...
-description: A list of tools available out-of-the-box on Codemagic macOS Apple silicon build machines using Xcode version 15.1.
-title: macOS Apple silicon build machine specification (Xcode 15.1+)
+title: Xcode 15.1.x
 aliases:
 
-weight: 13
+weight: 109
 ---
 
+Codemagic offers multiple build machines with different specifications and pre-installed tools. You can choose between them by specifying the required Xcode version.
```
helinanever, line 8: Here and elsewhere, it would make sense to mention instance/machine type as another variable. We can link to https://docs.codemagic.io/yaml-basic-configuration/yaml-getting-started/#instance-type
>You can choose between them by specifying the required Xcode version and instance type.

helinanever:
I'd also like to see the names unified in the menu, we can remove "specification" from Linux and Windows
image

helinanever:
One downside of having the Xcode pages in a collapsible menu category is the way the menu behaves when clicking on page -- the menu scrolls up, and in order to check another Xcode page, you'd have to scroll back down again. This seems to be problem with collapsible menu categories in general in our docs page and would be nice to have it fixed.

content/specs-macos/xcode-15-1.md
mikhail-tokarev: I don't think it's a good idea - we should keep focus on macOS and general available instance type.

helinanever (approved):
LGTM

PR #2853, opened 2024-11-26
bump Hugo version

mikhail-tokarev:
subj

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -208,7 +208,6 @@ mediaTypes:
 title: Codemagic Docs
 disableKinds:
     - taxonomy
-    - taxonomyTerm
     - section
 
 deployment:
--- a/layouts/partials/opengraph.html
+++ b/layouts/partials/opengraph.html
@@ -74,6 +74,6 @@
 {{ end }}
 
 {{/* Facebook Page Admin ID for Domain Insights */}}
-{{ with site.Social.facebook_admin }}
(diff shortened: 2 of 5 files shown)


config.yaml
```diff
@@ -208,7 +208,6 @@ mediaTypes:
 title: Codemagic Docs
 disableKinds:
     - taxonomy
-    - taxonomyTerm
```
mikhail-tokarev, line 211: > WARN  DEPRECATED: Kind "taxonomyterm" used in disableKinds is deprecated, use "taxonomy" instead.

layouts/partials/opengraph.html
```diff
@@ -74,6 +74,6 @@
 {{ end }}
 
 {{/* Facebook Page Admin ID for Domain Insights */}}
-{{ with site.Social.facebook_admin }}
+{{ with site.Params.Social.facebook_admin }}
```
mikhail-tokarev, line 77: > ERROR deprecated: .Site.Social was deprecated in Hugo v0.124.0 and will be removed in Hugo 0.139.0. Implement taxonomy 'social' or use .Site.Params.Social instead.

artemii-yanushevskyi (approved):
LGTM

Let's use 0.139.2 and add instructions to install Hugo with brew:

    brew install Hugo

I made sure it works locally

PR #2866, opened 2024-12-04
CodePush updates promoting downstream

icarusdust:
After testing an update against a deployment channel, it is possible to promote it to destination channels


kevin-suhajda (approved):
LGTM

PR #2868, opened 2024-12-04
update the Jave heap space error workaround with additional error

dtrdic:
Users started reporting an error which is similar to Java heap space
`Gradle build daemon has been stopped: since the JVM garbage collector is thrashing`
The workaround is the same as for Java heap space error

image


icarusdust (approved):
LGTM

PR #2879, opened 2024-12-09
Refactoring ios code signing documentation

dtrdic:
split automatic and manual codesigning to different pages


icarusdust:
We will review this as part of a larger doc updates strategy

PR #2908, opened 2025-01-07
Google Play release name field

mohammedbabelly20:
Add the new Google Play field `release_name`

### DIFF
--- a/content/yaml-publishing/google-play.md
+++ b/content/yaml-publishing/google-play.md
@@ -86,6 +86,10 @@ publishing:
     # or your custom track name
     track: internal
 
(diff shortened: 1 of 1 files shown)


content/yaml-publishing/google-play.md
```diff
@@ -86,6 +86,10 @@ publishing:
     # or your custom track name
     track: internal
 
+    # Optional text. The release name.
```
helinanever, line 89: seeing how we often specify the expected type of the value, then I'd go with "Optional string" here

helinanever (approved):

PR #2930, opened 2025-01-17
rotate deployment key

mikhail-tokarev:
every 90 days we rotate deployment key. This PR remove hardcoded key ID to easy rotate it via GitHub UI


helinanever (approved):
LGTM

garciaerubiel42-cmyk (commented):
that was interesting will fall back to the rest repo.

PR #2946, opened 2025-01-27
Enterprise plan update

icarusdust:
* Replacing M2 with M4 for base plan
* Adding an option with M4 Pro machines


kevin-suhajda (approved):
LGTM

PR #2990, opened 2025-02-20
Add section about token generation

helinanever:
Instead of linking to the Shorebird codemagic.yaml guide, provide instructions for creating the token in documentation.


dtrdic (approved):
looks good to me

PR #2998, opened 2025-02-26
Add missing environment block

ramit-current:
Added environment block with `google_credentials` because that is needed for the `GCLOUD_STORAGE_KEY` environment var to become available in the publish script


icarusdust:
Thanks for opening the PR, as this is a post-publishing script, I will add a reference link to the groups section instead in the description.

PR #3120, opened 2025-05-28
app stuck on splash screen

dtrdic:
Sometimes the app builds and deploys successfully to Firebase/Google play, but when downloaded and launched it remain stuck on the splash screen
image


icarusdust (approved):
LGTM

PR #3125, opened 2025-05-30
Correct Python version requirement

helinanever:
CLI tools require Python 3.8 as minimum version.


priitlatt (approved):
This is correct: Reference as of current latest version 0.58.3: https://github.com/codemagic-ci-cd/cli-tools/blob/4595f16ad336cfd2674c6e00ef36f6844b6f4930/pyproject.toml#L37-L38.

PR #3193, opened 2025-07-30
CLIENT_ID placeholder

icarusdust:
When defining a `label` value as an $environment variable, the variable needs to be defined inside `environment` in YAML separately with a placeholder, so it can pass the YAML validation


kevin-suhajda (approved):
LGTM

PR #3278, opened 2025-09-26
Make the placing of the `cache` section more concrete in `codemagic.yaml`

stewemetal:
If I undersand correctly, the `cache`  section should be added as a top-level section in the `codemagic.yaml` file. If that's the case, I think it would be nice to clarify that.

If it's not the case, then I believe the document should explicitly state that instead.

### DIFF
--- a/content/knowledge-codemagic/caching.md
+++ b/content/knowledge-codemagic/caching.md
@@ -32,13 +32,16 @@ A great article on Unity caching can be found in [our blog](https://blog.codemag
 
 {{< tab header="codemagic.yaml" >}}
 {{<markdown>}}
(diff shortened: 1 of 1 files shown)


stewemetal:
Reading the start of the doc again, there _is_ a hint that caching belongs to workflows,
> When you have enabled caching for a workflow, Codemagic will automatically generate a cache based on the output of the first successful build.
So I guess `cache` sections should be added to exact workflows then, and `cache` is not a global configuration. Please let me know which cas it true and I'll be happy to update the PR accordingly.
Cheers!

icarusdust (approved):
lgtm

PR #3280, opened 2025-09-29
Ms teams integration

dtrdic:
image

image


icarusdust:
it'be good to include a sample to post binaries as well besides messages

icarusdust (approved):
🚀

PR #3316, opened 2025-11-03
Scheme "xxx" not found from repository! Please reconfigure your project.

dtrdic:
image


icarusdust:
Maybe let's add other suggestions as well e.g. check if the scheme name is correct and if shared

PR #3370, opened 2026-01-08
Update link for Codemagic guide in Shorebird documentation

danagbemava-nc:
The original link appears to have changed on the shorebird side. Updated to the correct link now


helinanever (approved):
LGTM

PR #3377, opened 2026-01-21
Fix pricing explanation for Burstable concurrency usage

ivek1337:
Typo in Burstable concurrency pricing explanation "If you subscribe to 20 concurrencies and consume 10 within the 95th percentile, then you pay full price for 10 and 1/3 of the price for the remaining 10 (was 20 before)."


kevin-suhajda (approved):
LGTM

PR #3386, opened 2026-02-09
Update faqs

Becca-Saka:
new POC format for common issues page


icarusdust:
@Becca-Saka, check how it will look with other pages. and see if all the common issues have the same format - description and solution. Common iOS issues page, for example have `Cause` section as well.

PR #3429, opened 2026-04-01
Sibling VMs

icarusdust:
* Adding to highlight how the shared pool could cause performance issues 
* Pointing out how dedicated pools can play a role


remarkov (approved):
I agree with @helinanever, we may consider specifying that we may use **up to two** VMs on a single host during the busy time, otherwise LGTM.

PR #3443, opened 2026-04-15
Codepush analytics page adjustment

zgoldie:
I figured it would be best for the Analytics page to specifically mention the options of getting stats via the UI, CLI or API. Plus, I've added a mention of CodePush into the RN quickstart.

### DIFF
--- a/content/rn-codepush/codepush-analytics.md
+++ b/content/rn-codepush/codepush-analytics.md
@@ -23,9 +23,7 @@ Each project is listed below with its latest release and per-release download, i
 
 ### CLI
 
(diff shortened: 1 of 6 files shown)


zgoldie:
@icarusdust an idea for a couple of potential improvements

zgoldie:
Thinking about it more, maybe CodePush should be mentioned before the build step on the RN quickstart instead of at the end, since you don't need to run a build to do an OTA update. The /ci-integration page could also mention that a full binary build isn't needed, and that it can just make the JS bundle while running on a linux machine.

icarusdust:
will merge it in a new PR

PR #3453, opened 2026-04-30
Update CodePush pricing and access key management

helinanever:
- Add pricing info for CodePush PAYG
- Add section about access key generation


content/rn-codepush/setup.md
```diff
@@ -38,9 +38,9 @@ code-push --version
 
 This command prints the installed CLI version. If the installation was successful, you will see a version number (e.g., x.x.x). If the command is not found, it usually means the CLI was not installed correctly.
 
-## Get an access token
+## Authenticate the CLI
 
-The CodePush CLI authenticates using access tokens provided by the Codemagic team. Request a token [here](https://codemagic.io/contact-sales/).
+The CodePush CLI authenticates using access keys provided by Codemagic. Read more about how to obtain the access key [here](./security-and-access/#obtaining-the-access-key).
```
zgoldie, line 43: I reckon give the instructions here as well so the page is more self contained

To generate an access key, navigate to **OTA Updates > Manage Access Keys** and click **Generate key**. Enter an identifier for your key and select an expiration period. Once generated, the key is displayed only once — copy and store it securely before closing the dialog.

content/rn-codepush/security-and-access.md
```diff
@@ -58,6 +56,16 @@ This allows automated pipelines to publish updates without requiring an interact
 
 For configuring tokens in CI workflows, see [CI integration](/rn-codepush/ci-integration/). For initial setup and CLI authentication, see [Setup](/rn-codepush/setup/).
 
+### Obtaining the access key
+
+Users with an active CodePush subscription can manage access keys in the Codemagic application.
+
+To generate an access key, navigate to **OTA Updates > Manage Access Keys** and click **Generate key**. Enter an identifier for your key and select an expiration period. Once generated, the key is displayed only once — copy and store it securely before closing the dialog.
```
zgoldie: Can we hyperlink to that manage keys page?

helinanever: Unfortunately not, the URL contains team ID.

content/rn-codepush/setup.md
helinanever: That would only work for customers that already have an active CodePush subscription and can access the UI. Explaining the whole thing here again would create duplicate content. We actually link to the `/setup` page from the app as well, we could just as well assume that the people looking to authenticate are users that have already created an access key.

content/rn-codepush/security-and-access.md
mart-nc: Direction `OTA Updates > Manage Access Keys` makes it sound like "Manage Access Keys" is a setting/section somewhere on the page, while it's a button on the header. Maybe this could be improved?

mart-nc (approved):

PR #3455, opened 2026-05-06
App dome integration

todah-zg:
Please check if I got everything right on the Codemagic side. I couldn't test this in practice but as far as the AppDome is concerned, we are doing everything through their well documented Python SDK so that part should be ok.


kevin-suhajda (approved):
LGTM

PR #3456, opened 2026-05-07
A dedicated guide for migrating from Expo OTA to CodePush

todah-zg:
As far as I know, everything is correct. But I am not a RN developer or an EAS user so mistakes are possible.
If everything turns out to be correct, there is still a decision to be made on where to put this. Right now, it will appear as the last article in the 'React Native CodePush' section.
Note: there is also an older article 'Migrating from AppCenter' currently sitting with 'Quick start guides'. It makes sense to keep both of them together. And since CodePush is its own product, I personally prefer to keep them under CodePush umbrella so they don't get lost among hundreds of other non OTA related articles.

### DIFF
--- a/content/rn-codepush/migrating-from-expo-ota.md
+++ b/content/rn-codepush/migrating-from-expo-ota.md
@@ -1,0 +1,495 @@
+---
+title: 'Migrating from Expo Updates'
+description: Step-by-step guide to replacing Expo Updates (EAS Update) with Codemagic CodePush in a React Native app
+meta_title: Migrate from Expo Updates (EAS Update) to Codemagic CodePush
+meta_description: Learn how to migrate your React Native app from Expo Updates and EAS Update to Codemagic CodePush — including SDK swap, native configuration, CI pipeline changes, and a full concept mapping table.
+weight: 20
+---
+
+This guide walks through migrating a React Native app's OTA update mechanism from Expo Updates / EAS Update to Codemagic CodePush. It covers the conceptual differences between the two systems, maps equivalent concepts, and provides step-by-step instructions for completing the migration.
+
+---
+
+## Why migrate?
+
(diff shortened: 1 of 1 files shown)


content/rn-codepush/migrating-from-expo-ota.md
```diff
@@ -0,0 +1,495 @@
...
+| Channel (e.g. `preview`, `production`) | Deployment (e.g. `Staging`, `Production`) | Same idea; CodePush calls them deployments |
+| Branch | Deployment | EAS branches map roughly to CodePush deployments |
+| `eas update --channel production` | `code-push release-react MyApp-iOS ios` | Both bundle and publish JS in one command |
+| Runtime version | Target binary version (`--target-binary-version`) | Controls which native binary receives a given update |
+| `eas update:configure` | `code-push app add` + native file edits | CodePush configuration is done in native files |
+| `expo-updates` package | `@code-push-next/react-native-code-push` | The client-side SDK |
+| `updates.url` in `app.json` | `CodePushServerURL` in `Info.plist` / `strings.xml` | The update server endpoint |
+| EAS access token | CodePush access token | Used for CLI authentication |
+| `eas update --branch ...` promote | `code-push promote MyApp Staging Production` | Promotes a validated update to production |
+| Rollout (EAS doesn't have native %) | `--rollout 25` on `release-react` or `patch` | CodePush supports percentage-based rollouts |
```
kmsbernard: EAS Update does support percentage rollouts: https://docs.expo.dev/eas-update/rollouts/

todah-zg: Changed the last line to:
| `--rollout-percentage=10` | `--rollout 25` on `release-react` or `patch` | Specifies a percentage of users that should receive a new update |

kmsbernard (approved):

kevin-suhajda (approved):
LGTM

PR #3458, opened 2026-05-07
Agents learn from docs with technical excellence

todah-zg:
## Changelog:
- automatically generate llms.txt --> short intro + map to the whole site
- auto generate llms-full.txt --> dumps the entire site into a single markdown file. Apparently agents love this.
- all pages accessible as pure markdown by appending '.md' to the original url


layouts/_default/index.llmsfull.txt
```diff
@@ -0,0 +1,16 @@
...
+description: {{ . }}
+{{ end -}}
+last_modified: {{ .Lastmod.Format "2006-01-02" }}
+================================================================================
+
+# {{ .Title }}
+{{ with .Params.description }}
+> {{ . }}
+{{ end }}
+{{ .RawContent }}
```
AstDaria, line 15: `.RawContent `includes unrendered Hugo shortcodes (e.g., {`{< tabpane >}}, {{< tab >}}`), so LLMs will see this noise instead of the actual content. It’s better to use `.Plain`, which returns rendered, tag-stripped text — clean and readable.

AstDaria (changes requested):

layouts/_default/index.llmsfull.txt
todah-zg: Stripping shortcodes does simplify regular text but it completely wrecks code examples. Yaml workflows become one very long line without formatting. Please reconsider as I think AI can still manage shortcodes and would probably be better served by clear code examples.

AstDaria (approved):

PR #3469, opened 2026-05-15
Suggest skill

zgoldie:
Add a callout into the Setup page about installing the skill. Cursor thought that a `notebox` was the best approach, which works except having a nice bash box inside doesn't work hence the inline code.

image
<br>
<br>

I also added a quick mention on the Issues page, since that feels like a place where it's worth pointing out.

image


kevin-suhajda (approved):
LGTM

PR #3478, opened 2026-05-21
Fan - out / fan - in method of running builds

todah-zg:
Really proud of this one. It's a new way of orchestrating builds that meets an actual need we've seen (Banco de Credito del Peru but have heard this question from other prospects as well) without having to change anything in Codemagic itself.

Everything was thoroughly checked and double checked (Gemini is good at verifying Claude code and catching little inaccuracies) in theory. Did not have the chance for hands on testing. It does look believable, though.


kevin-suhajda (approved):
LGTM

PR #3488, opened 2026-06-01
Ai improvements

todah-zg:
- remove 404.md page from showing up in the Sitemap and in the llms.txt
- add Sitemap url to robots.txt
- remove /partials pages from appearing in the Sitemap
- llms.txt page section names now use more user friendly names instead of Hugo folder names

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -4,4 +4,4 @@ Allow: /
 User-agent: Twitterbot
 Allow: /
 
-Sitemap: {{ .Site.BaseURL }}sitemap.xml
+https://docs.codemagic.io/sitemap.xml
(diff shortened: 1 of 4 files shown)


static/robots.txt
```diff
@@ -3,3 +3,5 @@ Allow: /
 
 User-agent: Twitterbot
 Allow: /
+
+Sitemap: {{ .Site.BaseURL }}sitemap.xml
```
AstDaria: `Sitemap: {{ .Site.BaseURL }}sitemap.xml `won’t work — Hugo copies the `static/ `folder verbatim and only renders templates from `layouts/`, so this will output the literal `{{ .Site.BaseURL }}` (verified with a local build).

This will work instead: `https://docs.codemagic.io/sitemap.xml`

todah-zg: Ok. Changed

AstDaria (approved):

PR #3490, opened 2026-06-04
Update CodePush login command with URL

kevin-suhajda:
Add server URL to login request


kevin-suhajda:
From CLI version 1.4 the server URL is automatically added.

PR #3516, opened 2026-07-31
updates to the docs search issue

icarusdust:
- Section heading (subtitle) is now the primary title in results; page title shown below as secondary
  - Support contact message addition
  - Index updated to codemagic_docs_main
  - Rewrote the Algolia index template to split each page into per-section records by ## headings, so each section gets its own search result with the heading as subtitle and page title as title

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -141,7 +141,15 @@ const getResultHtml = (algoliaResultList, query) => {
         return null
     }
 
-    const results = algoliaResultList.map((result) => {
+    const preference = window.localStorage.getItem('preferred-configuration') ?? 'yaml'
+    const getTabScore = (uri) => {
+        if (preference === 'yaml' && (uri.startsWith('/yaml') || uri.startsWith('/rn-codepush'))) return 0
+        if (preference === 'flutter' && uri.startsWith('/flutter')) return 0
+        return 1
+    }
+    const sortedResults = algoliaResultList.slice().sort((a, b) => getTabScore(a.uri) - getTabScore(b.uri))
+
+    const results = sortedResults.map((result) => {
         const subtitle = result._highlightResult?.subtitle?.value ?? ''
         const pageTitle = result._highlightResult?.title?.value ?? result.title ?? ''
         return createHtmlElement('li', null, [
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -125,19 +125,6 @@ const getResultHtml = (algoliaResultList, query) => {
             innerText: 'Invalid search query: ' + algoliaResultList.message,
         })
 
-    const preferredConfiguration = localStorage.getItem('preferred-configuration') || defaultPreferredConfiguration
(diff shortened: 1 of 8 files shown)


assets/js/search.js
```diff
@@ -138,38 +138,50 @@ const getResultHtml = (algoliaResultList, query) => {
...
+            innerHTML: "Don't see what you're looking for? <a href='https://codemagic.io/contact/'>Contact us</a> or let us know via support chat widget in <a href='https://codemagic.io/apps'>app</a>.",
+        }),
+    ])
+
+    return createHtmlElement('ul', null, [...results, supportLink])
 }
 
 const getResults = (query) =>
-    query
+    query && query.trim().length >= 3 && !['and', 'the', 'then', 'or', 'but', 'for', 'with', 'not', 'from', 'this', 'that', 'are', 'was', 'has', 'how', 'what', 'when', 'where', 'who', 'why', 'can', 'use', 'using', 'set', 'get', 'its', 'into', 'also', 'other', 'next', 'used'].includes(query.trim().toLowerCase())
```
kerma: @mart-nc maybe want's to comment on how this could be made a bit more readable.

layouts/_default/index.json
```diff
@@ -1,17 +1,57 @@
...
+      {{- $text := replaceRE "(?s)```[^`]*```" "" $section }}
+      {{- $text = replaceRE "(?s)\\{\\{<\\s*highlight[^>]*>\\}\\}.*?\\{\\{<\\s*/highlight\\s*>\\}\\}" "" $text }}
+      {{- $text = $text | plainify | htmlUnescape }}
+      {{- $text = replaceRE `\{\{[^}]*\}\}` "" $text }}
+      {{- $text = replaceRE `#{1,6} ` "" $text }}
+      {{- $text = replaceRE `\*{1,3}([^*]+)\*{1,3}` "$1" $text }}
+      {{- $text = replaceRE `_{1,3}([^_]+)_{1,3}` "$1" $text }}
+      {{- $text = replaceRE "`[^`]+`" "" $text }}
+      {{- $text = replaceRE `!?\[([^\]]*)\]\([^)]*\)` "$1" $text }}
+      {{- $text = replaceRE `\s+` " " $text | strings.TrimSpace }}
```
kerma: What's the goal for these replacements? It's a pretty difficult piece of code to maintain, I'm wondering whether the are better ways to achieve what you're after here.

icarusdust: Its main idea is that instead of indexing each page as a single flat record, it splits each page into per-section records, one record per ##heading

Refactored it by removing duplicates and adding a single loop

icarusdust:
Added badges to search results. Results are now grouped by the tab where the search was initiated: results from the current tab are listed first, followed by results from the other tab.

mart-nc (approved):

kerma (approved):

PR #3524, opened 2026-08-17
Codemagic init

icarusdust:
Adding `codemagic init`, explaining what it does and how to use it


helinanever (approved):
Docs look good to me

PR #3532, opened 2026-08-26
Update starting builds automatically

icarusdust:
Pointing out automatic webhook addition when working with WFE


helinanever (approved):
LGTM

PR #3543, opened 2026-09-07
Pushing people to Patch

zgoldie:
The decision was to keep the CodePush docs for people already using it, but nudge people towards Patch. So a few changes for that are:

- A note at the top of the codepush setup page, which is the main codepush page people look at
image
</p>

- Pointing people to Patch in the RN quickstart instead of Codepush
image
</p>

- Having Patch as the OTA solution in the homepage, not CodePush
image


assets/scss/_article.scss
zgoldie, line 1: The info card and h3 within it were rendering weirdly, hence these changes

icarusdust:
what about also adding legacy? e.g. React Native CodePush (Legacy)

icarusdust (approved):

